PR #12, opened 2019-08-01
Cl generic compose extrinsic

clangenb:
* PoC Generic Extrinsic that is table to call Balance::transfer with Custom MyCall enum. To make it dynamic and truly generic, metadata parsing needs to be implemented.
* Added AccountKey enum, which is able to hold an ed25519 or an sr25519 key pair and providing signing methods for it. While composing extrinsic, the key type to be used is passed as an argument.


substrate-bip39/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "substrate-bip39"
+version = "0.3.1"
+authors = ["Parity Technologies <admin@parity.io>"]
+license = "GPL-3.0"
+edition = "2018"
+description = "Converting BIP39 entropy to valid Substrate (sr25519) SecretKeys"
```
electronix: Should we indicate somewhere that this version 0.3.1 has modifications and is not the same as the "official" 0.3.1? From the perspective of a user who could simply say: "Oh, it's the same version, let's take the official"

clangenb: done in cl-auto-config-from-metadata

src/extrinsic/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+// 			signature: None,
+// 			function: xt.function,
+// 		},
+// 	}
+// }
+
+#[cfg(test)]
+mod tests {
+	use node_primitives::Balance;
+	use node_runtime::{BalancesCall, Call};
```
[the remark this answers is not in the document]
clangenb: done

PR #24, opened 2019-09-19
Api client tutorial

clangenb:
- small tutorial code including documentation md.

- reading flow still needs to be improved. But let me know what you think before I dwell on details too much.


brenzi:
overall, this looks very nice and accessible. Even some learnings about substrate in general.

electronix:
Comments from @brenzi fixed with my commits.

Issue #26, opened 2019-09-21
drop dependency of examples on test node_runtime crate


brenzi:
It is cumbersome to fix dependency issues because we import substrate-test-runtime

1. the build script enforces wasm build which is not necessary here
1. too many dependencies are added for only a fow type definitions

brenzi:
done, except `example_contracts.rs` which depends on runtime's `Event` type. Can we define that ourselves? If I use `node_runtime::Event` it won't recognize a contract Event - obviously.

https://github.com/scs/substrate-api-client/tree/no_std_rstd

clangenb:
We can do that. The event enum is constructed in the same manner as the call enums, which can therefore be constructed from the metadata. I reckon approximately 2-3 ours to have that work. 

Would do that with a macro in order handle the dynamic argument types.

brenzi:
Unfortunately, the root cause why we have to disable the contract example right now is still an open issue: 
https://github.com/rust-lang/cargo/issues/6571

So, if we can work around the Event enum dependency easily, that'd be great

Issue #16, opened 2019-08-28
compose_extrinsic needs additional dependencies


brenzi:
I'm trying to use this lib within my own crate (example with generic extrinsic). The following occurs:

```

error[E0432]: unresolved import `codec`
  --> src/lib.rs:38:14
   |
38 |       let xt = compose_extrinsic!(
   |  ______________^
39 | |         api.clone(),
40 | |         "Balances",
41 | |         "transfer",
42 | |         GenericAddress::from(to),
43 | |         Compact(42 as u128)
44 | |     );
   | |_____^ use of undeclared type or module `codec`
   |
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

...
error[E0433]: failed to resolve: use of undeclared type or module `primitives`
...
error[E0432]: unresolved import `log`
```

all of which probably caused by these imports:
https://github.com/scs/substrate-api-client/blob/9e33717bcc2546806ce7201f4772aaa9da3e58f8/src/extrinsic/mod.rs#L68

brenzi:
a comment would be nice that the following dependencies should be added to Cargo.toml:
```
log = "0.4"
env_logger = "0.6"
...
[dependencies.codec]
default-features = false
package = "parity-scale-codec"
features = ["derive"]
version = "1.0.0"

[dependencies.primitives]
git = 'https://github.com/paritytech/substrate'
rev = '6d47fd919b4d86e4348c6c19d99c80372587d215'
package = "substrate-primitives"
```

PR #28, opened 2019-09-22
No std rstd

brenzi:
This adds `no_std` support. (without depending on `sgx_tstd`)
In order for this to work, I had to fork `substrate-primitives` and enable crypto primitives.
This also drops dependency on custom runtime crate.

TODO:
* `example_contract` is broken. blocker: #26


brenzi:
I suggest to merge although example_contract is broken. If someone wants to use it, I've added a hint

brenzi:
the recent merge commit broke no_std, need to fix this first

brenzi:
no_std fix works

src/extrinsic/mod.rs
```diff
@@ -22,8 +22,7 @@ pub mod balances;
 pub mod contract;
 pub mod xt_primitives;
 
-pub extern crate codec;
-pub extern crate log;
```
clangenb: Has this been done for a reason?

I made this for issue #16 . This re-exports the dependencies locally. Making them available in the macro via `crate::...`.

Then third parties using `substrate-api-client` do not need to import the `codec` and `log` dependencies

clangenb:
apart from my comment i think this is ready to merge.

src/extrinsic/mod.rs
brenzi: the reason for removing it is that it breaks `no_std`

clangenb: try reexporting conditionally, only if in `std` environment

PR #31, opened 2019-10-13
Cl drop account key

### LINKED ISSUE #27: drop DYI AccountKey use primitives::crypto::Pair instead
Now that we've ported `substrate-primitves` to `no_std` sgx-style providing crypto primitives as well, we should drop our cutom type `AccountKey` in favor of upstream `Pair`

See: 
https://github.com/scs/substrate-api-client/tree/no_std_rstd

including the ported primitves crate

clangenb:
- Drop crypto module from substrate-api-client. Instead, use substrate-primitives::crypto
- This implies that from now on the Api is generic with Api<P: crypto::Pair>
- the predefined extrinsics are now methods of Api instead pure functions.

Closes #27


src/examples/example_contract.rs
```diff
@@ -59,7 +61,7 @@ fn main() {
     let wasm = wabt::wat2wasm(CONTRACT).expect("invalid wabt");
 
     // 1. Put the contract code as a wasm blob on the chain
-    let xt = contract::put_code(api.clone(), 500_000, wasm);
+    let xt = api.put_code(500_000, wasm);
```
brenzi: I don't like the namespacing: put_code and create refer to contract module only but the get a top-level function in the api. I'd prefer `api.contract.put_code()` and `api.contract.create()`

src/examples/example_contract.rs
```diff
@@ -78,7 +80,7 @@ fn main() {
     println!("[+] Event was received. Got code hash: {:?}\n", code_hash);
 
     // 2. Create an actual instance of the contract
-    let xt = contract::create(api.clone(), 1_000, 500_000, code_hash, vec![1u8]);
+    let xt = api.create(1_000, 500_000, code_hash, vec![1u8]);
```
brenzi: see above

src/examples/example_contract.rs
```diff
@@ -101,7 +103,7 @@ fn main() {
     );
 
     // 3. Call the contract instance
-    let xt = contract::call(api.clone(), deployed_at, 500_000, 500_000, vec![1u8]);
+    let xt = api.call(deployed_at, 500_000, 500_000, vec![1u8]);
```
brenzi: see above

src/examples/example_contract.rs
clangenb: I have been trying to get my head arround this. We could do it like: https://github.com/paritytech/substrate-subxt/blob/master/src/srml/balances.rs

But I am not sure if this is unnecesarily complex. Maybe just revert it to the way it was?

brenzi: My suggestion is only about the naming, not the general approach. It could also be a prefix like àpi.balances_transfer()` and `api.contract_put_code()`. A touch less elegant though.

brenzi: let's leave this for a refactoring task. I don't want to keep this PR stuck because of a detail

PR #36, opened 2019-10-27
adding option to select json request id

brenzi:
allows lib user to specify id for each RPC subscription. necessary if there's more than one


brenzi:
maybe we should get rid of REQUEST_TRANSFER at the same time? Need to adapt example code

clangenb:
Looks fine. I agree that we should get rid to REQUEST_TRANSFER. Which would need a change in the [client](https://github.com/scs/substrate-api-client/blob/43124b582d5bbc42e8147d7080c59a21051483a9/src/rpc/client.rs) as well.

PR #39, opened 2019-11-01
add sudo example

### LINKED ISSUE #38: support sudo calls
this client should support sudo calls. I've started some ugly hack which doesn't work yet:
https://github.com/scs/substrate-api-client/tree/brenzi-sudo

brenzi:
closes #38 
new sudo extrinsic example
added low-level getters as well


clangenb:
Ah interesting, this will be helpfull for many.

PR #40, opened 2019-11-01
Jenkins integration

electronix:


clangenb:
Nice! I haven't taken a thorough look yet. From my general impression though, would you think it makes sense to squash some commits to get a better overview of the feature beeing introduced?

electronix:
You are perfectly right. Just wanted to test the PR feature with this. Will do this immediately afterwards

PR #43, opened 2019-11-06
adding pub fn for double map storage

brenzi:
this is tested against encointer-node.
There is a caveat:
the runtime module has to define the double_map with blake2_256 hasher, not twox128. We could add functions for both, but shawns tutorial silently assumes blake2

this works:
```
decl_storage! {
	trait Store for Module<T: Trait> as EncointerCeremonies {
		ParticipantRegistry get(participant_registry): double_map CeremonyIndexType, blake2_256(ParticipantIndexType) => T::AccountId;
	}
}
```

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -25,7 +25,7 @@ fn storage_key_hash_vec(module: &str, storage_key_name: &str, param: Option<Vec<
     let mut key = [module, storage_key_name].join(" ").as_bytes().to_vec();
     match param {
         Some(par) => {
-            key.append(&mut par.clone());
+            key.extend(&par.clone());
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -25,7 +25,7 @@ fn storage_key_hash_vec(module: &str, storage_key_name: &str, param: Option<Vec<
     let mut key = [module, storage_key_name].join(" ").as_bytes().to_vec();
     match param {
         Some(par) => {
-            key.append(&mut par.clone());
+            key.extend(&par.clone());
```
clangenb: Tiny detail; I believe `par` is cloned twice as extend does implicitly apply clone.

PR #48, opened 2019-12-17
support wasm

alanpoon:
Breaking changes:
- Added support for async/ await
- At the moment, no intention to support for subscribe_events function
- The websocket result is able to be read by nodejs and react using wasmpack
- Change std::mpsc::sync::channel to future signals
- Perform a quick hack on future signals to resolve https://github.com/Pauan/rust-signals/issues/14

```rust
//Rust node_module
#[wasm_bindgen]
pub async fn k(url:String)->String{
    let mut api:Api<Pair> = Api::new(format!("ws://{}", url)).await;
    // get some plain storage value
    let result_str = api.get_storage("Balances", "TotalIssuance", None).await.unwrap();
    let result = hexstr_to_u256(result_str).unwrap();
    result
}
```

In javascript:
```javascript
const wasm = await import('rust'); //your rust node_module
const resy =  await wasm.k("127.0.0.1:9944");
console.log("HHHHHHHHHHHHHHHHHH",resy);
```


brenzi:
@alanpoon thanks for this PR. As you're introducing breaking changes we won't be able to merge this before we made sure all our downstream projects can be easily migrated. We haven't even introduced semver so far. 
This may take some time...

S3bb1:
@alanpoon i wanted to use your branch to extend it to the current implementation. I'm now made it compile but stuck with your example 
`
//Rust node_module
#[wasm_bindgen]
pub async fn k(url:String)->String{
    let mut api:Api<Pair> = Api::new(format!("ws://{}", url)).await;
    // get some plain storage value
    let result_str = api.get_storage("Balances", "TotalIssuance", None).await.unwrap();
    let result = hexstr_to_u256(result_str).unwrap();
    result
}
`

Do you have a working full example of it?

alanpoon:
> @alanpoon i wanted to use your branch to extend it to the current implementation. I'm now made it compile but stuck with your example
> 
> ```rust
> //Rust node_module
> #[wasm_bindgen]
> pub async fn k(url:String)->String{
>     let mut api:Api<Pair> = Api::new(format!("ws://{}", url)).await;
>     // get some plain storage value
>     let result_str = api.get_storage("Balances", "TotalIssuance", None).await.unwrap();
>     let result = hexstr_to_u256(result_str).unwrap();
>     result
> }
> ```
> 
> Do you have a working full example of it?

You can look at this https://github.com/alanpoon/belanja_app/tree/developer
-There is shell script: https://github.com/alanpoon/belanja_app/blob/developer/wasm.sh
(my wasm-bindgen version is 0.2.47)
-package.json https://github.com/alanpoon/belanja_app/blob/developer/package.json#L92

-webpack.config https://github.com/alanpoon/belanja_app/blob/developer/webpack.config.js
-implementation: https://github.com/alanpoon/belanja_app/blob/developer/screens/WasmComponent.js

haerdib:
Closing this, as we now support wasm. Thanks a lot for the PR and your Input!

PR #54, opened 2020-02-23
send multiple extrinsics without waiting for finalization

### LINKED ISSUE #53: send extrinsics without waiting for finalization
In order to send a high frequency of xt we need a non-blocking way to send extrinsics.

caveat: It's not enough to return after the "ready" msg. The problem is that the nonce needs to increase before the xt is included and the nonce is updated on-chain

brenzi:
fixes #53 

no_std not building yet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,11 @@ package = "sp-keyring"
 version = "2.33"
 features = ["yaml"] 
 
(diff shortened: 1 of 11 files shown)


brenzi:
upstream fix should be included in 2.0.0pre.alpha.4

PR #60, opened 2020-03-23
added chain_api: chain_get_header, chaing_get_block, chain_get_finalized_head

### LINKED ISSUE #50: get block api
'''rust
pub fn chain_get_block(hash:String) -> Value {
    json!({
    "method": "chain_getBlock",
    "params": [hash],
    "jsonrpc": "2.0",
    "id":"1"
    })
}
'''
Any support for rpc chain_getBlock?

clangenb:
* Extended rpc api
* closes #50

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,7 +53,6 @@ pub mod extrinsic;
 #[cfg(feature = "std")]
 pub mod node_metadata;
 
-#[cfg(feature = "std")]
 pub mod utils;
 #[cfg(feature = "std")]
 pub mod rpc;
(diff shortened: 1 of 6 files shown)


clangenb:
ci fails as it tries to run the example against a more recent node

src/lib.rs
```diff
@@ -204,6 +203,18 @@ impl<P> Api<P>
         hexstr_to_u256(result_str).unwrap()
     }
 
+    pub fn get_finalized_head(&self) -> WsResult<String> {
```
brenzi: String result isn't very nice. can we decode this to `Header`? I guess we could use the generic header in `use sr_primitives::generic::Header;`

brenzi:
```
let ret = api.get_header(None).unwrap();
let hdr: Header = serde_json::from_str(&ret).unwrap();
```

brenzi:
included in PR #66

PR #62, opened 2020-04-01
V2.0.0 alpha.5

### LINKED ISSUE #61: upgrade to v2.0.0-alpha.5
this is rather involved. We need to support generic hashers, as defined in metadata:
```
       {
        "name": "Account",
        "modifier": "Default",
        "ty": {
         "Map": {
          "hasher": "Blake2_128Concat",
          "key": "T::AccountId",
          "value": "AccountInfo<T::Index, T::AccountData>",
          "unused": false
         }
        },

```

And for double_map, the two hashers can be different

```
       {
        "name": "MeetupRegistry",
        "modifier": "Default",
        "ty": {
         "DoubleMap": {
          "hasher": "Blake2_256",
          "key1": "CurrencyCeremony",
          "key2": "MeetupIndexType",
          "value": "Vec<T::AccountId>",
          "key2_hasher": "Blake2_256"
         }
        },

```

brenzi:
fixes #61 

* now building against a [fork of substrate-node-template](https://github.com/scs/substrate-node-template/tree/test-api-client)
* storage map and double map working. extended example 
* minor refactoring
* works for v2.0.0-alpha.5 based [encointer-client](https://github.com/encointer/encointer-node/tree/master/client)

### TODO before merging!
* fix all examples!
* node template has no kitties and no contracts. do we really want to maintain substrate-test-nodes? We could just maintain a branch on a fork of node-template, like I started to do (see above)

### TODO maybe later
* unsure, if using AccountId32 is a good idea. I think we should use the Runtime itself a a type argument to Api, like subxt does. then we can derive `Runtime::AccountId`

### DIFF
--- a/src/examples/example_get_storage.rs
+++ b/src/examples/example_get_storage.rs
@@ -17,7 +17,6 @@
 use clap::{load_yaml, App};
 use codec::Encode;
 use keyring::AccountKeyring;
-
 use sp_core::crypto::Pair;
 use substrate_api_client::{utils::{hexstr_to_u256, hexstr_to_hash}, Api};
 
@@ -39,6 +38,13 @@ fn main() {
         .unwrap();
     let result = hexstr_to_hash(result_str);
     println!("[+] block hash for blocknumber 42 is {:?}", result); 
+
+    // get StorageDoubleMap
+    let result_str = api
+        .get_storage_double_map("TemplateModule", "SomeDoubleMap", 1_u32.encode(), 2_u32.encode())
+        .unwrap();
+    let result = hexstr_to_u256(result_str);
+    println!("[+] some double map (1,2) should be 3. Is {:?}", result); 
 
     // get Alice's AccountNonce with api.get_nonce()
     let signer = AccountKeyring::Alice.pair();
--- a/src/examples/example_get_storage.rs
+++ b/src/examples/example_get_storage.rs
@@ -18,7 +18,7 @@ use clap::{load_yaml, App};
 use codec::Encode;
 use keyring::AccountKeyring;
 use sp_core::crypto::Pair;
-use substrate_api_client::{utils::{hexstr_to_u256, hexstr_to_hash}, Api};
+use substrate_api_client::{utils::{hexstr_to_u256, hexstr_to_hash}, Api, Hash};
 
 fn main() {
     env_logger::init();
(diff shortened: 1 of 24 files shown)


src/examples/example_get_storage.rs
```diff
@@ -28,27 +27,24 @@ fn main() {
...
+    // get StorageMap
     let accountid = AccountKeyring::Alice.to_account_id();
-    let result_str = api
-        .get_storage("System", "AccountNonce", Some(accountid.encode()))
+    let result: Hash = api
+        .get_storage_map("System", "BlockHash", 1u32).or(Some(Hash::default()))
         .unwrap();
-    let result = hexstr_to_u256(result_str).unwrap();
-    println!("[+] Alice's Account Nonce is {}", result.low_u32());
+    println!("[+] block hash for blocknumber 42 is {:?}", result);
```
clangenb, line 38: Honestly, storage map key `1u32` for the `block_hash` resolves to blocknumber 42?? Nice

src/lib.rs
```diff
@@ -197,51 +179,113 @@ where
...
-                arr.clone_from_slice(key.to_owned().public().as_ref());
-                Ok(Self::_get_nonce(self.url.clone(), arr))
+                arr.clone_from_slice(pair.to_owned().public().as_ref());
+                let accountid: AccountId = Decode::decode(&mut &arr.encode()[..]).unwrap();
+                if let Some(info) = self.get_account_info(&accountid) {
+                    Ok(info.nonce)
+                } else { Ok(0) }
             }
-            None => Err("Can't get nonce when no signer is set"),
+            None => Err("Can't get )nonce when no signer is set"),
```
clangenb, line 190: typo

src/lib.rs
```diff
@@ -197,62 +183,144 @@ where
...
+        self.get_storage_by_key_hash(storagekey.0)
+    }
+
+    pub fn get_account_data(&self, address: &AccountId) -> Option<AccountData> {
+        if let Some(info) = self.get_account_info(address) {
+            Some(info.data)
+        } else { None }
+    }
+
+    pub fn get_finalized_head(&self) -> WsResult<String> {
```
clangenb, line 216: I agree, that WsResult<String> is ugly. We better make concrete types

src/lib.rs
```diff
@@ -197,62 +183,144 @@ where
...
-        storage_prefix: &str,
-        storage_key_name: &str,
-        param: Option<Vec<u8>>,
-    ) -> WsResult<String> {
-        Self::_get_storage(self.url.clone(), storage_prefix, storage_key_name, param)
+        storage_prefix: &'static str,
+        storage_key_name: &'static str,
+    ) -> Option<V> {
+        let storagekey: sp_core::storage::StorageKey = self.metadata
+                .module(storage_prefix).unwrap()
```
clangenb, line 238: Although this is not critial code, I am a little bit concerned about these `unwrap()`s. 

As they all return `MetadataError`,  would recommend to make the function return: `Result<V, MetadataError>` and use `module()?`

src/lib.rs
```diff
@@ -197,62 +183,144 @@ where
...
     }
 
-    pub fn send_extrinsic(&self, xthex_prefixed: String) -> WsResult<Hash> {
+    pub fn get_storage_by_key_hash<V:Decode+Clone>(&self, hash: Vec<u8>) -> Option<V> {
+        let mut keyhash_str = hex::encode(hash);
+        keyhash_str.insert_str(0, "0x");
+        let jsonreq = json_req::state_get_storage(&keyhash_str);
+        if let Ok(hexstr) = Self::_get_request(self.url.clone(), jsonreq.to_string()) {
+            info!("storage hex = {}", hexstr);
+            let hexstr = hexstr
```
clangenb, line 267: Maybe we should generify the `utils::hexstr_vec` function to be able to call it here?

`pub fn decode_from_hexstr<V: Decode>(hexstr: String) -> Option<V>/Result<V, FromHexError>`

brenzi:
The question is if an uninitialized storage should lead to an error or not. Either way, we need to treat the case with `if let` or `match`.  To me, `Option` is more intuitive, because either there is a value or there isn't....Is there some idiomatic best practise for this?

clangenb:
I agree that an `Option` is intuitive if a storage is there or not. But retrieving a storage might fail somewhere else, i.e., if a module is not there or if the decoding fails. Currently, we just `unwrap()` in most cases.

From a third party developper's perspective, he might want to handle himself if the process fails instead of having the api-client just panic.

I have seen code that returns a `Result<Option<T>, ..>`, maybe something like that makes sense?

PR #64, opened 2020-04-13
add Identity default hasher

alekspickle:
Worked for me. *shrug*


brenzi:
identity hasher is already included in #62. Will be merged soonish

PR #66, opened 2020-04-24
V2.0.0 alpha.6

### LINKED ISSUE #61: upgrade to v2.0.0-alpha.5

brenzi:
* upgrading
* deprecating hard-to-maintain samples

fixes #61 

* now building against a [fork of substrate-node-template](https://github.com/scs/substrate-node-template/tree/test-api-client)
* storage map and double map working. extended example 
* minor refactoring
* includes new api for get_block aso.
* works for v2.0.0-alpha.5 based [encointer-client](https://github.com/encointer/encointer-node/tree/master/client)

### technical debt
* node template has no kitties and no contracts. do we really want to maintain substrate-test-nodes? We could just maintain a branch on a fork of node-template, like I started to do (see above)
* unsure, if using AccountId32 is a good idea. I think we should use the Runtime itself a a type argument to Api, like subxt does. then we can derive `Runtime::AccountId`


brenzi:
@clangenb please check if this PR includes all your previous work

src/extrinsic/xt_primitives.rs
```diff
@@ -21,12 +21,16 @@ use sp_std::prelude::*;
...
+use sp_core::H256;
 use sp_core::blake2_256;
 use sp_runtime::{generic::Era, MultiSignature};
-pub type GenericAddress = Address<AccountId, AccountIndex>;
+
+pub use sp_runtime::AccountId32 as AccountId;
+
+pub type AccountIndex = u64;
+
+pub type GenericAddress = AccountId; //Address<AccountId, AccountIndex>;
```
clangenb, line 33: why do that redundant rename? I think it would be better to use AccountId

src/lib.rs
```diff
@@ -101,7 +117,7 @@ where
 
         let meta = Self::_get_metadata(url.clone());
         let metadata = Metadata::try_from(meta).unwrap();
-        info!("Metadata: {:?}", metadata);
+        debug!("Metadata: {:?}", metadata);
```
clangenb, line 120: really happy about that!

src/lib.rs
```diff
@@ -197,62 +181,143 @@ where
...
-                arr.clone_from_slice(key.to_owned().public().as_ref());
-                Ok(Self::_get_nonce(self.url.clone(), arr))
+                arr.clone_from_slice(pair.to_owned().public().as_ref());
+                let accountid: AccountId = Decode::decode(&mut &arr.encode()[..]).unwrap();
+                if let Some(info) = self.get_account_info(&accountid) {
+                    Ok(info.nonce)
+                } else { Ok(0) }
             }
-            None => Err("Can't get nonce when no signer is set"),
+            None => Err("Can't get )nonce when no signer is set"),
```
clangenb, line 192: typo

src/node_metadata.rs
```diff
@@ -265,12 +320,14 @@ impl<K: Encode, V: Decode + Clone> StorageMap<K, V> {
         bytes.extend(&sp_core::twox_128(&self.storage_prefix)[..]);
         let encoded_key = key.encode();
         let hash = match self.hasher {
+            StorageHasher::Identity => encoded_key.to_vec(),
             StorageHasher::Blake2_128 => sp_core::blake2_128(&encoded_key).to_vec(),
             StorageHasher::Blake2_128Concat => {
                 // copied from substrate Blake2_128Concat::hash since StorageHasher is not public
-                sp_core::blake2_128(&encoded_key)
+                let x: &[u8] = &encoded_key[..];
```
clangenb, line 327: you could as well use `encoded_key.as_slice()` , which I personally prefer

src/node_metadata.rs
```diff
@@ -288,6 +345,49 @@ impl<K: Encode, V: Decode + Clone> StorageMap<K, V> {
...
+    pub fn key(&self, key1: K, key2: Q) -> StorageKey {
+        let mut bytes = sp_core::twox_128(&self.module_prefix).to_vec();
+        bytes.extend(&sp_core::twox_128(&self.storage_prefix)[..]);
+        for (encoded_key, hasher) in [(key1.encode(), self.hasher.clone()), (key2.encode(), self.key2_hasher.clone())].iter() {
+            let hash = match hasher {
+                StorageHasher::Identity => encoded_key.to_vec(),
+                StorageHasher::Blake2_128 => sp_core::blake2_128(&encoded_key).to_vec(),
+                StorageHasher::Blake2_128Concat => {
+                    // copied from substrate Blake2_128Concat::hash since StorageHasher is not public
+                    let x: &[u8] = &encoded_key[..];
```
clangenb, line 369: why make x?

clangenb:
and yes, it does include my work

src/extrinsic/xt_primitives.rs
brenzi: yes, but the problem is that we will want to get back to generic in order to support polkadot/kusama  chains as well which have a different definition.

src/node_metadata.rs
brenzi: ok. was copy-paste from upstream, but I agree

src/node_metadata.rs
brenzi: ???
anyway. will replace this redundant code by a fn call

Issue #65, opened 2020-04-15
await InBlock or Broadcast status when sending extrinsic


brenzi:
(for v2.0.0-alpha.5 and later)
The lifetime of an extrinsic goes like this:
```
send {"id":"3","jsonrpc":"2.0","method":"author_submitAndWatchExtrinsic","params":["0x2d0284d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d017693d988ff9b4d3577c0b9d2b9424e647fbec5875fef096644e311cb7499fd2a7cb99850c7e2806ffafecb3c63de17caa382340752077d6ca63e8995564f7d820068000400ce8f99245b836fa93f021c02c14a80d5029e2d43fb6265a0c610e5ac8d043e1b02286bee"]}
get {"jsonrpc":"2.0","result":232,"id":"3"}
get {"jsonrpc":"2.0","method":"author_extrinsicUpdate","params":{"result":"ready","subscription":232}}
get {"jsonrpc":"2.0","method":"author_extrinsicUpdate","params":{"result":{"broadcast":["QmfSF4VYWNqNf5KYHpDEdY8Rt1nPUgSkMweDkYzhSWirGY","Qmchhx9SRFeNvqjUK4ZVQ9jH4zhARFkutf9KhbbAmZWBLx","QmQJAqr98EF1X3YfjVKNwQUG9RryqX4Hv33RqGChbz3Ncg"]},"subscription":232}}
get {"jsonrpc":"2.0","method":"author_extrinsicUpdate","params":{"result":{"inBlock":"0x3104d362365ff5ddb61845e1de441b56c6722e94c1aee362f8aa8ba75bd7a3aa"},"subscription":232}}
get {"jsonrpc":"2.0","method":"author_extrinsicUpdate","params":{"result":{"finalized":"0x3104d362365ff5ddb61845e1de441b56c6722e94c1aee362f8aa8ba75bd7a3aa"},"subscription":232}}
```
currently, we only offer to submit and await `ready` or `finalized`.
It would make sense to support `inBlock` too and maybe `broadcast`

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 70.0 DAI (70.0 USD @ $1.0/DAI)  attached to it as part of the web3foundation fund.__

[...]

PR #69, opened 2020-04-30
add: Await InBlock and  Broadcast Status When Sending Extrinsic

developerfred:
#65 

### update jenkins
https://github.com/scs/substrate-api-client/pull/74

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,7 +344,19 @@ where
                 info!("ready: {}", res);
                 Ok(None)
             }
-            _ => panic!("can only wait for finalized or ready extrinsic status"),
+            XtStatus::Broadcast => {
+                rpc::send_extrinsic(self.url.clone(), jsonreq, result_in);
+                let res = result_out.recv().unwrap();
+                info!("broadcast: {}", res);
+                Ok(None)
+            }
+            XtStatus::InBlock => {
+                rpc::send_extrinsic(self.url.clone(), jsonreq, result_in);
+                let res = result_out.recv().unwrap();
+                info!("inBlock: {}", res);
+                Ok(Some(hexstr_to_hash(res).unwrap()))
+            }
+             _ => panic!("can only wait for finalized, ready, inBlock or broadcast extrinsic status"),
         }
     }
 
--- a/src/rpc/client.rs
+++ b/src/rpc/client.rs
@@ -22,6 +22,8 @@ use ws::{CloseCode, Handler, Handshake, Message, Result, Sender};
 #[derive(Debug, PartialEq)]
 pub enum XtStatus {
     Finalized,
+    InBlock,
+    Broadcast,
     Ready,
     Future,
     Error,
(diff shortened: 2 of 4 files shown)


brenzi:
@developerfred thanks for your PR. May I ask you to fix the unit test?

developerfred:
@brenzi I will fix

brenzi:
It could be that the error actually comes from testing against the wrong node in jenkins. The api-client should now actually test against our node-template #71

developerfred:
@brenzi I'll take a look at the issue #71

brenzi:
@developerfred I'v manually tested your PR. It builds, but it doesn't work. Here's how to reproduce:

```
https://github.com/scs/substrate-node-template.git
git checkout test-api-client
cargo build --release
./target/release/node-template purge-chain --dev
./target/release/node-template --dev

# other terminal
https://github.com/developerfred/substrate-api-client.git
git checkout await-InBlock 
cargo build --release --examples
RUST_LOG=debug ./target/release/examples/example_transfer
```
The log should show you that the client waits for `finalized`

now, change the example:
```
diff --git a/src/examples/example_transfer.rs b/src/examples/example_transfer.rs
index 094cbf6..4ae73e5 100755
--- a/src/examples/example_transfer.rs
+++ b/src/examples/example_transfer.rs
@@ -47,7 +47,7 @@ fn main() {
 
     // send and watch extrinsic until finalized
     let tx_hash = api
-        .send_extrinsic(xt.hex_encode(), XtStatus::Finalized)
+        .send_extrinsic(xt.hex_encode(), XtStatus::InBlock)
         .unwrap();
     println!("[+] Transaction got finalized. Hash: {:?}\n", tx_hash);
```
I would expect it to wait for `InBlock` now, but it only awaits `Ready`.
That's because you're calling the exact same function that we wrote to wait for `ready`. I'll comment in your commit

src/lib.rs
```diff
@@ -344,6 +344,18 @@ where
                 info!("ready: {}", res);
                 Ok(None)
             }
+            XtStatus::Broadcast => {
+                rpc::send_extrinsic(self.url.clone(), jsonreq, result_in);
```
brenzi, line 348: you can't call this function because that is hard-coded to wait for `ready`.

src/lib.rs
```diff
@@ -344,6 +344,18 @@ where
...
+                let res = result_out.recv().unwrap();
+                info!("broadcast: {}", res);
+                Ok(None)
+            }
+            XtStatus::InBlock => {
+                rpc::send_extrinsic(self.url.clone(), jsonreq, result_in);
```
brenzi, line 354: same here

src/lib.rs
developerfred: I will add a new await function

src/lib.rs
developerfred: I will add a new await function

developerfred:
@brenzi fixed pull on https://github.com/scs/substrate-api-client/commit/934414657436f507df53ba07b6e33978968d7798

src/rpc/client.rs
```diff
@@ -150,7 +189,16 @@ fn parse_status(msg: &str) -> (XtStatus, Option<String>) {
...
                     (XtStatus::Finalized, Some(hash.to_string()))
-                } else {
+                }
+                else if let Some(hash) = obj.get("inBlock") {
+                    info!("inBlock: {:?}", hash);
+                    (XtStatus::InBlock, Some(hash.to_string()))
+                }
+                else if let Some(hash) = obj.get("broadcast") {
+                    info!("Broadcast: {:?}", hash);
+                    (XtStatus::Broadcast, Some(hash.to_string()))                     (XtStatus::InBlock, Some(hash.to_string()))
```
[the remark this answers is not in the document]
developerfred: sorry @brenzi !

brenzi:
I'm sorry that I have to disappoint you @developerfred but we do prefer to wait for a task until it is well done and reviewing it when it is tested and ready

Issue #68, opened 2020-04-30
Getting hex`s OddLength error constructing extrinsic


alekspickle:
Reproduce is kind of hard because of our architecture but I can privide `compose_extrinsic` macro input to check.

```
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn check_compose_extrinsics() {
        let node_url = "wss://node1-chain.akropolis.io";
        let api = Api::new(node_url.into()).set_signer(get_sr25519_pair(ANY_VALID_MNEMONIC));
        let from = H160::default(); // any eth address would suffice
        let to = AccountId32::default();
        let id = H256::default();
        let ext = compose_extrinsic!(
            api,
            "Bridge",
            "multi_signed_mint",
            id,
            from,
            GenericAddress::from(to),
            Compact(0u32),
            Compact(10000000000000000000u128)
        );
        let hash = api.send_extrinsic(ext.hex_encode(), XtStatus::Finalized);
        //                      ^^^^^^^^^^^^^^^^^---- fails here
    }
}

pub fn get_sr25519_pair(signer_mnemonic_phrase: &str) -> sr25519::Pair {
    sr25519::Pair::from_phrase(signer_mnemonic_phrase, None)
        .expect("invalid mnemonic phrase")
        .0
}
```

upd: I reproduced this as a test so it should be fairly easy

developerfred:
@alekspickle do you want to have an error handling in the function?

alekspickle:
@developerfred 
In theory that should not be possible, I assume?
You added an odd check which returns `""` if it is odd, but how is that even possible from compose_extrinsics macro?+

developerfred:
@alekspickle thanks, I will update this pr

alekspickle:
Just debugged to the point ehere I can say that this error is rising because of node error handling you use. 
`hex::decode` fails to parse "nix" that you substitute [here](https://github.com/scs/substrate-api-client/blob/23b6255502141559fef42fafd7032704c5dd1f46/src/rpc/client.rs#L126).
Since `None` value is returned for each error from node error parsing [if the value of result is neither an object nor the null](https://github.com/scs/substrate-api-client/blob/23b6255502141559fef42fafd7032704c5dd1f46/src/rpc/client.rs#L154) the result field that you can get in some cases is the reason of substitution of hexed input for `hex::decode` with `"nix"`.

[...]

PR #70, opened 2020-05-01
fix: OddLength error constructing extrinsic

developerfred:
#68


brenzi:
Hey @developerfred. Thanks for your PR.

`hex::encode` should never return an odd-length result. The cause of #68 must be something else

PR #73, opened 2020-05-01
change the way to parse errors

alekspickle:
As I [stated](https://github.com/scs/substrate-api-client/issues/68#issuecomment-622381438) in #68 earlier, treating all rpc errors like `None` and substituting it with "nix" is very mutch error-prone. 
Passing the error on the other hand behaves similarly but makes debug easier.
So in case of error, [this log](https://github.com/scs/substrate-api-client/blob/23b6255502141559fef42fafd7032704c5dd1f46/src/lib.rs#L338) will give you exact error from node.
ps. the error in test is real btw


brenzi:
very nice, thank you!

Issue #71, opened 2020-05-01
fix jenkins, test against substrate-node-template


brenzi:
We have dropped maintenance of substrate-test-nodes and test against a modified substrate-node-template now:

https://github.com/scs/substrate-node-template/tree/test-api-client

This should be integrated for Jenkins to succeed

PR #74, opened 2020-05-01
fix: jenkins, test against substrate-node-template

developerfred:
#71


brenzi:
Thanks for this @developerfred, but please don't work on issues that were already assigned to someone. As you don't have access to our CI you can't test it anyway....

PR #82, opened 2020-05-11
use substrate-api-client-test-node

### LINKED ISSUE #71: fix jenkins, test against substrate-node-template

electronix:
- Use the new test-node
- Update of Jenkinsfile
  - Copy first because of file permissions
  - Comment sudo example as it generates an error

This fixes #71

### DIFF
--- a/.githooks/README.md
+++ b/.githooks/README.md
@@ -1,0 +1,3 @@
+# githooks
+
+Copy these files to `.git/hooks`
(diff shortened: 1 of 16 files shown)


brenzi:
@electronix finally!!!, please squash-merge!

Issue #88, opened 2020-05-19
Mortal Transactions?


whalelephant:
I would like to see if there are plans to support mortal transactions?

brenzi:
no plans for this. But we're glad to accept a PR if you'd like to implement it

PR #91, opened 2020-05-22
Allow for immortal transactions in compose_extrinsic_offline macro

whalelephant:
PR referencing Issue #88 

- mortal era for compose_extrinsic_offline macro
- example_benchmark_bulk_xt updated to use macro
- example_compose_extrinsic_offline updated to use macro

### DIFF
--- a/src/extrinsic/mod.rs
+++ b/src/extrinsic/mod.rs
@@ -68,8 +68,8 @@ macro_rules! compose_extrinsic_offline {
     $genesis_hash: expr,
     $genesis_or_current_hash: expr,
     $runtime_spec_version: expr) => {{
-        use $crate::extrinsic::xt_primitives::*;
         use sp_runtime::generic::Era;
+        use $crate::extrinsic::xt_primitives::*;
         let extra = GenericExtra::new($era, $nonce);
         let raw_payload = SignedPayload::from_raw(
             $call.clone(),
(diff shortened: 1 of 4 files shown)


src/extrinsic/mod.rs
```diff
@@ -56,25 +56,28 @@ macro_rules! compose_call {
...
     $call: expr,
     $nonce: expr,
+    $era: expr,
     $genesis_hash: expr,
+    $genesis_or_current_hash: expr,
     $runtime_spec_version: expr) => {{
         use $crate::extrinsic::xt_primitives::*;
-
-        let extra = GenericExtra::new($nonce);
+        use sp_runtime::generic::Era;
```
clangenb, line 72: this is a tiny rustfmt error.

whalelephant:
Thanks! resolved. 

I was not sure if `compose_extrinsic_offline!` or `compose_extrinsic!` would be best for immortal era to be included, ended up with this way because it was less change to the code. Happy to discuss more.

clangenb:
Yes, I have been thinking too if we should make the `Era` available as a parameter to all the functions that generate an extrinsic as an `Option<Era>`, or add a macro called `compose_mortal_extrinsic`. I would go for the latter and I would not reject a pull request that adds the macro.

But for now I think this is fine as it introduces the least code bloat. I am unsure about the use cases of mortal transactions. On our side we have (obviously) never used them.

PR #94, opened 2020-06-04
make less verbose

clangenb:
replace several `log::info!` with `log::debug!`


clangenb:
I have been playing around with this a little bit and had to decide the less verbose model is not optimal either. Thus, closing this one.

PR #95, opened 2020-06-06
get map key prefix

whalelephant:
This may be useful for getting keys with key prefix in storage maps

- use `state_getKeys` in json_req
- add `get_storage_map_key_prefix` method for api client


clangenb:
Hi Whalelefant,

Thanks for your pull request! Looks good an useful to me. Will merge that one 👍

PR #105, opened 2020-11-18
optimize: throw error to user instead panic

AurevoirXavier:


brenzi:
thanks for this. I'll have a look ASAP

AurevoirXavier:
Btw, why you send this req twice?

https://github.com/scs/substrate-api-client/blob/master/src/lib.rs#L352

AurevoirXavier:
This lib doesn't handle any error, so painful...

brenzi:
I agree that there is potential regarding error handling. Currently we lack funding for proper maintenance of this lib, so we're only fixing critical stuff. This may change soon

src/lib.rs
```diff
@@ -337,7 +332,7 @@ where
         &self,
         key: StorageKey,
         at_block: Option<Hash>,
-    ) -> Option<V> {
+    ) -> WsResult<V> {
```
brenzi, line 335: this is problematic. There are basically two things that can go wrong: communication and storage lookup.
A WsResult should indicate a communication error but we also need the `Option` to indicate if the storage key has yielded an entry or not. so, probably, this should actually be `WsResult<Option<V>>`

brenzi (changes requested):
Some examples are broken and need to be fixed. I'd suggest to be verbose in examples to show how the error handling is done. I.e. for examples_get_storage we could do:

```
    // get StorageDoubleMap
    let result: u32 = match api
        .get_storage_double_map("TemplateModule", "SomeDoubleMap", 1_u32, 2_u32, None) {
            Ok(ret) => match ret {
                Some(v) => v,
                None => panic!("storage is empty")
            },
            Err(e) => panic!("fetching SomeDoubleMap failed: {}", e)
    };

```


If you are willing to fix these issues, thanks a lot! If not, please redirect your PR to the following branch so we can fix the issues ourselves: 
https://github.com/scs/substrate-api-client/tree/error-handling

brenzi:
@AurevoirXavier Thanks a lot for your PR. We have chosen a slightly different approach which is now merged in #107. We hope this covers your requirement for error handling.

PR #107, opened 2020-12-14
Error handling

clangenb:
Introduce proper error handling and propagate errors to top level.

This changes the api. Allmost all methods return a `Result<T, ApiClientError>` now where T mostly is of type `Option<T>`

+ added `FromHexTrait`
+ added `api.signer_account(&self) -> Option<AccountId32>`


brenzi:
this looks nice and clean and also covers #105. examples and tutorials covered.
There seems to be a CI issue

@electronix can you have a look why CI fails?

electronix:
I will

PR #118, opened 2021-05-24
Add GitHub action

clangenb:


clangenb:
I close this. I will introduce this directly with another PR. I thought that GA needs to have the file in master already to be able to activate its hooks. Apparently, it doesn't.

PR #127, opened 2021-06-30
fix twox64concat hasher

haerdib:


haerdib:
Ups, there is still a bug. Converting it to draft for now

src/node_metadata.rs
```diff
@@ -635,6 +635,12 @@ fn key_hash<K: Encode>(key: &K, hasher: &StorageHasher) -> Vec<u8> {
         StorageHasher::Blake2_256 => sp_core::blake2_256(&encoded_key).to_vec(),
         StorageHasher::Twox128 => sp_core::twox_128(&encoded_key).to_vec(),
         StorageHasher::Twox256 => sp_core::twox_256(&encoded_key).to_vec(),
-        StorageHasher::Twox64Concat => sp_core::twox_64(&encoded_key).to_vec(),
+        StorageHasher::Twox64Concat => {
+            sp_core::twox_64(bytes)
```
brenzi, line 639: where does "bytes" come from? compare to a few lines up. shouldn't it work the same way?

haerdib: Yup, that's  the bug - I was a bit a rash with the PR. Fix is incoming

haerdib: Fixed

PR #137, opened 2021-07-13
fix + remove full-crypto feature flag and make it implicit.

clangenb:
This removes the need to depend on `sp-application-crypto` in the depending crate  when the api-client is used in a no_std environment.


haerdib (approved):
LGTM

PR #142, opened 2021-07-21
Add staking extrinsics

trevor-crypto:
Overview:
- Added the most common (in my opinion) staking extrinsics
- Added api call for getting fee details (payment_queryFeeDetails)

~~The custom `FeeDetails` and `InclusionFee` are added because we would need both `pallet-transaction-payment` and `sp-rpc` (for `NumberOrHex`), which is quite a lot of extra stuff.~~

~~It also may not be necessary to add the entire `pallet-staking` crate, if we are ok with adding a copy of `RewardDestination`. I'm not sure if this pallet will be useful for anything else in the future. Please let me know and I will remove the import.~~

### DIFF
--- a/src/std/rpc/mod.rs
+++ b/src/std/rpc/mod.rs
@@ -18,6 +18,8 @@ use serde::{Deserialize, Serialize};
 
 #[cfg(feature = "ws-client")]
 pub use ws_client::{EventsError, WsRpcClient};
+
+use crate::Balance;
 #[cfg(feature = "ws-client")]
 mod ws_client;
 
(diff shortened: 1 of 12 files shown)


trevor-crypto:
Not able to see the CI server. Could someone let me know what the error is?

clangenb:
@haerdib can you check jenkins?

clangenb:
@trevor-crypto seems to be cargo clippy, which makes me realize that our GA CI does fail to fail upon clippy warnings. :D

trevor-crypto:
@clangenb Thanks! Sometimes they don't show up locally for me either :man_shrugging:

src/std/rpc/mod.rs
```diff
@@ -55,3 +57,59 @@ pub struct ReadProof<Hash> {
     /// A proof used to prove that storage entries are included in the storage trie
     pub proof: Vec<sp_core::Bytes>,
 }
+
+// Based on structures here
+// https://github.com/paritytech/substrate/blob/cf086eeb894e61b45c732b3dad2ce4dcc54fa1af/frame/transaction-payment/src/types.rs#L68
+// Would require pallet-transaction-payment and sp-rpc (for NumberOrHex) since the RPC returns either a number or hex string
+#[derive(Debug, Serialize, Deserialize)]
```
clangenb, line 64: Have you checked the amount of extra packages that are included? From my experience, the substrate crates have a lot of overlapping dependencies, so it might not matter that much if we include them.

clangenb (commented):
Thank you again for a very clean looking PR!

Regarding the staking-pallet. You are right, it pulls in quite some dependencies; hence I would prefer it to be feature-gated.

src/std/rpc/mod.rs
trevor-crypto: I just looked and there doesn't seem to be anything crazy...I will test it again on my end.

trevor-crypto: Done, with feature-gate for staking

clangenb:
Thanks for the PR. Looks very good to me. 

@haerdib can you merge this?

clangenb (approved):

PR #145, opened 2021-08-04
Change return type of Api::get_fee_details()

trevor-crypto:
Changed return type of get_fee_details() from FeeDetails<NumberOrHex> to
FeeDetails<Balance> to make it more usable.

### DIFF
--- a/src/std/mod.rs
+++ b/src/std/mod.rs
@@ -405,7 +405,11 @@ where
 }
 
 fn convert_fee_details(details: FeeDetails<NumberOrHex>) -> ApiResult<FeeDetails<u128>> {
-    let inclusion_fee = inclusion_fee_with_balance(details.inclusion_fee)?;
+    let inclusion_fee = if let Some(inclusion_fee) = details.inclusion_fee {
+        Some(inclusion_fee_with_balance(inclusion_fee)?)
+    } else {
+        None
+    };
     let tip = details
         .tip
         .try_into()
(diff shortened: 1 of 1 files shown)


src/std/mod.rs
```diff
@@ -399,6 +404,39 @@ where
...
+
+fn inclusion_fee_with_balance(
+    inclusion_fee: Option<InclusionFee<NumberOrHex>>,
+) -> ApiResult<Option<InclusionFee<Balance>>> {
+    let inclusion_fee = if let Some(inclusion_fee) = inclusion_fee {
+        Some(InclusionFee {
+            base_fee: inclusion_fee
+                .base_fee
+                .try_into()
+                .map_err(|_| ApiClientError::TryFromIntError)?,
```
trevor-crypto, line 424: Awful, due to the real `TryFromIntError` not implementing `Debug` (it's only a `()`!) , so I can't add it properly to `ApiClientError`

clangenb: Hahah, I feel you.

src/std/mod.rs
```diff
@@ -399,6 +404,39 @@ where
...
+    let inclusion_fee = inclusion_fee_with_balance(details.inclusion_fee)?;
+    let tip = details
+        .tip
+        .try_into()
+        .map_err(|_| ApiClientError::TryFromIntError)?;
+    Ok(FeeDetails { inclusion_fee, tip })
+}
+
+fn inclusion_fee_with_balance(
+    inclusion_fee: Option<InclusionFee<NumberOrHex>>,
```
clangenb, line 417: I think an Option as a sole argument does not make much sense. I think it is better to check on the outside if it is set with a `map_or_default` or something like that. What do you think?

clangenb (commented):
Thanks again for your PR, only one small comment.

src/std/mod.rs
trevor-crypto: Yeah, that's much better. Let me change it

trevor-crypto: Changed to if-let so I can `?` the result

clangenb: Yest this is fine with me, I just think it is good to see on top-level that nothing happens with a `None`.

One thing I came to like recently (but this is only a matter of style) would be:

`fee_opt.map(inclusion_fee_with_balance).transpose()?`

I don't think it needs to be changed, just letting you know because I think this pattern is not well known. :)

clangenb (approved):
Nice, looks good to me

clangenb:
@haerdib ready to merge from my point of view. I think you can decide whether you want to first merge this PR or your PR.

haerdib:
> 
> 
> @haerdib ready to merge from my point of view. I think you can decide whether you want to first merge this PR or your PR.

Lest merge this one first. But first I need to fix Jenkins. I will merge as soon as it has run through.

PR #152, opened 2021-08-23
add end_process to error response

### LINKED ISSUE #135: client does not return upon error message
The worker freezes when receving an error message from node:
For example:
`[2021-06-23T09:38:39Z ERROR ws::handler] WS Error <Custom(Extrinsic("extrinsic error code 1014: Priority is too low: (39323901909911507 vs 14018908424301368): The transaction has too low priority to replace another transaction already in the pool."))>`
or
`[2021-06-23T09:40:54Z ERROR ws::handler] WS Error <Custom(Extrinsic("extrinsic error code 1010: Invalid Transaction: Inability to pay some fees (e.g. account balance too low)"))>`

The error most probably lies here:
https://github.com/scs/substrate-api-client/blob/master/src/rpc/client.rs#L154
https://github.com/scs/substrate-api-client/blob/master/src/rpc/client.rs#L170

Upon Error the  receiver thread is not ended and https://github.com/scs/substrate-api-client/blob/master/src/lib.rs#L519 will never finish.

Moved from isse: https://github.com/integritee-network/worker/issues/300 (respective from https://github.com/Polkadex-Substrate/polkadexTEE-worker/issues/153)

haerdib:
Fixes #135 

Maybe there's a more elegant solution, but it does fix the issue.

### DIFF
--- a/src/std/rpc/ws_client/mod.rs
+++ b/src/std/rpc/ws_client/mod.rs
@@ -210,7 +210,7 @@ pub fn on_extrinsic_msg_until_finalized(
         Err(e) => {
             end_process(out, result, None)?;
             Err(Box::new(e).into())
-        },
+        }
         _ => Ok(()),
     }
 }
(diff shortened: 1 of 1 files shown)


clangenb (approved):
I think this approach is just fine.

Issue #138, opened 2021-07-13
return error details upon included but failed extrinsic


echevrier:
When the node send an error, we get only an ignoring unsupported module event. It would be good to have a nice way to get the error and listen to these errors

clangenb:
If I understand correctly, you want to listen to an event that is triggered when the extrinsic fails?

Listening to custom events is supported, you  can do it like as follows:

https://github.com/integritee-network/worker/blob/25c2b69d127d1998d8ed6b3cc3ce6cb98efd2228/client/src/main.rs#L562-L587

echevrier:
In this case we are waiting for the BlockConfirmed event. If the node throws an error and no BlockConfirmed event is sent, how can we catch the error?

haerdib:
If the node throws an error before the extrinsic Finalized, the api client will return the error message (if you're listening until finalisation, like it's done here: 
https://github.com/integritee-network/worker/blob/25c2b69d127d1998d8ed6b3cc3ce6cb98efd2228/client/src/main.rs#L416-L418

Does this answer your question?

echevrier:
Sorry I don't know. Here the description of my case:

I'm listening to a create event like this :
```rust
let tx_hash = chain_api
        .send_extrinsic(xt.hex_encode(), XtStatus::InBlock)
        .unwrap();
    //subscribe to event Created
    let (events_in, events_out) = channel();
    chain_api.subscribe_events(events_in).unwrap();

    //Wait for Created event to extract and return the NFTid
    let mut decoder = EventsDecoder::try_from(chain_api.metadata.clone()).unwrap();
    decoder.register_type_size::<NFTId>("NFTId").unwrap();
    decoder
        .register_type_size::<AccountId>("AccountId")
        .unwrap();
    decoder
        .register_type_size::<NFTSeriesId>("NFTSeriesId")
        .unwrap();

    let account_id = get_accountid_from_str(owner_ss58);
    debug!("AccountId of signer  {:?}", account_id);

    //For now no possibility to catch here the errors coming from chain. infinite loop.
    //See issue https://github.com/scs/substrate-api-client/issues/138#issuecomment-879733584
    loop {
        let ret = chain_api
            .wait_for_event::<CreatedArgs>("Nfts", "Created", Some(decoder.clone()), &events_out)
            .unwrap();

        info!("Created event received");
        if ret.account_id == account_id {
            return Some(ret.nft_id);
        }
    }   
```
[...]

PR #153, opened 2021-08-24
Fix issue #138 

echevrier:
and add example_event_error_details to test/show how to use the wait_for_event

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -67,6 +67,7 @@ pipeline {
         sh 'target/release/examples/example_get_storage'
         sh 'target/release/examples/example_get_blocks'
         sh 'target/release/examples/example_benchmark_bulk_xt'
+        sh 'target/release/examples//example_event_error_details'
         // TODO: example needs fixing sh 'target/release/examples/example_sudo'
       }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -67,7 +67,7 @@ pipeline {
         sh 'target/release/examples/example_get_storage'
         sh 'target/release/examples/example_get_blocks'
         sh 'target/release/examples/example_benchmark_bulk_xt'
-        sh 'target/release/examples//example_event_error_details'
+        sh 'target/release/examples/example_event_error_details'
         // TODO: example needs fixing sh 'target/release/examples/example_sudo'
       }
     }
--- a/src/examples/example_event_error_details.rs
+++ b/src/examples/example_event_error_details.rs
@@ -32,10 +32,17 @@ fn main() {
         .unwrap();
 
     let from_account_id = AccountKeyring::Alice.to_account_id();
-    match api.get_account_data(&from_account_id).unwrap() {
-        Some(alice) => println!("[+] Alice's Free Balance is is {}\n", alice.free),
-        None => println!("[+] Alice's Free Balance is is 0\n"),
-    }
+
+    let amount = match api.get_account_data(&from_account_id).unwrap() {
+        Some(alice) => {
+            println!("[+] Alice's Free Balance is is {}\n", alice.free);
+            alice.free
+        }
+        None => {
+            println!("[+] Alice's Free Balance is is 0\n");
+            10000000000000000000
+        }
+    };
 
     let to = AccountKeyring::Bob.to_account_id();
 
@@ -44,7 +51,6 @@ fn main() {
         None => println!("[+] Bob's Free Balance is is 0\n"),
     }
     // generate extrinsic
-    let amount = 10000000000000000000;
     let xt = api.balance_transfer(MultiAddress::Id(to.clone()), amount);
 
     println!(
@@ -60,13 +66,14 @@ fn main() {
         .unwrap();
     println!("[+] Transaction got included. Hash: {:?}\n", tx_hash);
 
+    //Transfer will failed as Alice want to transfer all her balance. She has not enough money to pay the fee
     let (events_in, events_out) = channel();
     api.subscribe_events(events_in).unwrap();
     let args: ApiResult<TransferEventArgs> =
         api.wait_for_event("Balances", "Transfer", None, &events_out);
     match args {
         Ok(transfer_event) => {
-            println!("Error : Transfer event received!!!\n");
+            println!("Transfer event received!!!\n");
             println!("Transactor: {:?}", transfer_event.from);
             println!("Destination: {:?}", transfer_event.to);
             println!("Value: {:?}", transfer_event.value);
--- a/src/examples/example_event_error_details.rs
+++ b/src/examples/example_event_error_details.rs
@@ -1,0 +1,100 @@
+use std::sync::mpsc::channel;
+
+use clap::{load_yaml, App};
+use codec::Decode;
+use keyring::AccountKeyring;
+use sp_core::crypto::Pair;
+use sp_runtime::AccountId32 as AccountId;
+use sp_runtime::MultiAddress;
+
+use substrate_api_client::rpc::WsRpcClient;
+use substrate_api_client::sp_runtime::app_crypto::sp_core::sr25519;
+use substrate_api_client::{Api, ApiResult, XtStatus};
+
+// Look at the how the transfer event looks like in in the metadata
+#[derive(Decode)]
+struct TransferEventArgs {
+    from: AccountId,
+    to: AccountId,
+    value: u128,
+}
+
+fn main() {
+    env_logger::init();
+    let url = get_node_url_from_cli();
+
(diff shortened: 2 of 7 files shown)


haerdib:
@echevrier jenkins failed due to a cargo fmt mismatch: 
Diff in /var/lib/jenkins/workspace/aTEE_substrate-api-client_PR-153/src/std/rpc/ws_client/events.rs at line 210

src/std/rpc/ws_client/events.rs
```diff
@@ -1,246 +1,263 @@
-// Copyright 2019 Parity Technologies (UK) Ltd.
```
clangenb, line 1: Do we have a line-ending issue here?

src/examples/example_event_error_details.rs
```diff
@@ -0,0 +1,100 @@
...
+        .unwrap();
+    println!("[+] Transaction got included. Hash: {:?}\n", tx_hash);
+
+    let (events_in, events_out) = channel();
+    api.subscribe_events(events_in).unwrap();
+    let args: ApiResult<TransferEventArgs> =
+        api.wait_for_event("Balances", "Transfer", None, &events_out);
+    match args {
+        Ok(transfer_event) => {
+            println!("Error : Transfer event received!!!\n");
```
clangenb, line 69: Is the error here correct?

clangenb (commented):
Cool, looks very nice and clean. But I believe we need to fix that line issue problem there.

src/examples/example_event_error_details.rs
echevrier: It should not get there, so it is in error if this happens

src/std/rpc/ws_client/events.rs
echevrier: What shall I check? I ran cargo fmt and I don't understand why it happens in this file only.

haerdib: Check out https://www.jetbrains.com/help/clion/configuring-line-endings-and-line-separators.html, maybe that can help

Jenkinsfile
```diff
@@ -67,6 +67,7 @@ pipeline {
         sh 'target/release/examples/example_get_storage'
         sh 'target/release/examples/example_get_blocks'
         sh 'target/release/examples/example_benchmark_bulk_xt'
+        sh 'target/release/examples//example_event_error_details'
```
haerdib, line 70: there's one `/` too many in `examples//example_evexample_event_error_details`

src/examples/example_event_error_details.rs
```diff
@@ -0,0 +1,100 @@
...
+    let tx_hash = api
+        .send_extrinsic(xt.hex_encode(), XtStatus::InBlock)
+        .unwrap();
+    println!("[+] Transaction got included. Hash: {:?}\n", tx_hash);
+
+    let (events_in, events_out) = channel();
+    api.subscribe_events(events_in).unwrap();
+    let args: ApiResult<TransferEventArgs> =
+        api.wait_for_event("Balances", "Transfer", None, &events_out);
+    match args {
```
haerdib, line 67: Jenkins Output:
```log
+ target/release/examples//example_event_error_details
Interacting with node on ws://127.0.0.1:9944

[+] Alice's Free Balance is is 1152921453830787971

[+] Bob's Free Balance is is 1152921505007848018

Sending an extrinsic from Alice (Key = 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY),

to Bob (Key = 5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty)

[+] Composed extrinsic: UncheckedExtrinsic(Some((MultiAddress::Id(d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)), GenericExtra(Era::Immortal, 403, 0))), ([5, 0], MultiAddress::Id(8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 (5FHneW46...)), 1000000000000000000))

[+] Transaction got included. Hash: Some(0x07fa345ac583560385e3861dd82c41f785807ba577ac3bb88bd5a7fa8c8e21b9)

Error : Transfer event received!!!

Transactor: d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d (5GrwvaEF...)
Destination: 8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48 (5FHneW46...)
Value: 1000000000000000000
[+] Bob's Free Balance is now 2152921505105848018

[+] Alice's Free Balance is now 152921441357773710
```

is that expected?

src/examples/example_event_error_details.rs
clangenb: Ah, I see. We want the error here. :) This was not clear to me, maybe be more clear about that?

src/std/rpc/ws_client/events.rs
clangenb: To prevent further cargo fmt discrepancies, we should use a [rustfmt.toml](https://github.com/integritee-network/worker/blob/master/rustfmt.toml). This one defines the line-endings to be `Unix` style.

clangenb: But maybe you are right, and the file was wrong before. :)

src/examples/example_event_error_details.rs
haerdib: @clangenb I'd love to hear your opinion on this matter:

Jenkins uses the examples as tests against the node, however, most examples only print the output but don't actually test anything else than connecting to the node and receivin the events. So upon unexpected events (such as in this example ; the transfer event was successful) it fails to fail.

Are the examples here for the examples sake or should we modify them to actually test something? (ofc in a different issue, not in this one)

src/std/node_metadata.rs
```diff
@@ -269,6 +308,37 @@ impl ModuleWithEvents {
...
+        &self.name
+    }
+
+    pub fn error(&self, index: u8) -> Result<&String, MetadataError> {
+        self.errors
+            .get(&index)
+            .ok_or(MetadataError::ErrorNotFound(index))
+    }
+
+    pub fn print(&self) {
```
haerdib, line 329: Cool! Works neatly, thanks alot!

haerdib (changes requested):
Seems to work neatly, thanks! Only the example does not what it should do - atleast not according to the comments because the balance of Bob is changing. If you adapt the example, then I'm more than happy.

src/examples/example_event_error_details.rs
echevrier: Sorry. The amount to transfer is not big enough, so the transfer coud be executed. I will increase the value.

src/examples/example_event_error_details.rs
echevrier: If it is an example, the error makes no sense. If it is a test, then we should improve the info message. For now I remove the Error in the message and we can update it if the examples shall test something. See comment from haerdib https://github.com/scs/substrate-api-client/pull/153#discussion_r695442582

src/examples/example_event_error_details.rs
clangenb: The primary purpose of the example is to verify that we can successfully talk to the node, and also to show how things are done. Implicitly, we have some testing because we `unwrap` many times. I have checked the other examples, and I believe that unwraps suffice.

Exact testing would require some effort, as the return value would depend on the state of the blockchain, which means we can only meaningfully test on a clean node, so you can't run the example twice on the same node. Additionally, some tests currently depend on each other, e.g,  Alice's nonce is continuously increasing.

src/std/rpc/ws_client/events.rs
echevrier: According to https://github.com/scs/substrate-api-client/pull/153?file-filters%5B%5D=.rs#discussion_r695428706, my settings are not set to unix line-endings. If I change them, all the files are affected. As the problem happened only in one file, I'm  not sure I have to do it

clangenb: Uff, it seems that you are right. We are currently not using unix-file endings. Then I suggest, we leave it as is here, and I will make a PRwith [this](https://github.com/scs/substrate-api-client/tree/cl-add-rustfmt-toml) branch, where I added the rustfmt.toml.

clangenb (approved):
From my point of view, we can merely merge this. 👍

haerdib (approved):
LGTM, Jenkins now prints the error:

`[2021-08-25T10:29:36Z ERROR substrate_api_client::std::rpc::ws_client] Extrinsic Failed: InsufficientBalance`

PR #164, opened 2021-09-19
upgrade substrate to 558a68f and drop StorageDoubleMap example

brenzi:
* upgrade to paritytech/substrate@558a68f
* dropped StorageDoubleMap in template and get_storage example. too painful to maintain


clangenb:
If we drop the double map example, we can also drop the dependency on the scs node, as the substrate-node-template suffices, no?

brenzi:
yes, we could drop that.... almost
node-template does not export SignedBlock, which is needed for the get-blocks example

clangenb:
Made an issue regarding `SignedBlock`:
* https://github.com/scs/substrate-api-client/issues/165

PR #168, opened 2021-09-28
update frame-metadata to new crate + cargo update

haerdib:


brenzi (approved):
I don't like hardcoding v13, but I guess we can lift that and become generic again once we upgrade beyond v14?

clangenb:
I am not certain if we can un-hardcode the metadata version. Before, substrate had only one version marked as valid and the rest as deprecated in the rust code. I don't believe they have changed that now.

clangenb (approved):
Cool. Looks good to me.

haerdib:
> 
> 
> I don't like hardcoding v13, but I guess we can lift that and become generic again once we upgrade beyond v14?

Yes - by default the newest version will be taken (see reexport : https://github.com/paritytech/frame-metadata/blob/main/frame-metadata/src/lib.rs#L49-L53) . So if we upgrade so v14 we will be able to remove the v- specification.

PR #169, opened 2021-10-01
Add `XtStatus::SubmitOnly`

clangenb:
I was having the impression that we get a certain lag when the enclave tries to submit extrinsics in the ocall.

So I tried to add `send_extrinsic` variant that is even faster than waiting for  `XtStatus::Ready`. I was having trouble finding out, what the exact checks are when you wait until ready. So, I just implemented it and wanted to test it.

Subjectively, I am not confident that it makes a difference. @brenzi, @haerdib  do you know if this is even worth implementing?

### DIFF
--- a/src/examples/example_transfer.rs
+++ b/src/examples/example_transfer.rs
@@ -52,7 +52,7 @@ fn main() {
 
     // send and watch extrinsic until finalized
     let tx_hash = api
-        .send_extrinsic(xt.hex_encode(), XtStatus::InBlock)
+        .send_extrinsic(xt.hex_encode(), XtStatus::SubmitOnly)
         .unwrap();
     println!("[+] Transaction got included. Hash: {:?}\n", tx_hash);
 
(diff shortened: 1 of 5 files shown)


clangenb:
Nevermint. The lag was completely unrelated. Stupid me has not seen that we still wait until the next block in said ocall in the worker. 

Regardless. My question remains.

haerdib (commented):
If already implemented I don't see a reason not to integrate it. Substrate offers this rpc call, so atleast they deemed it worth to be implemented, why should we not?
On first sight the PR looks clean enough, we might want to add a different test than to change the example_transfer, but lets wait for what @brenzi thinks on this one.

brenzi:
My knowledge my be a bit dated on this. On a real chain, `Broadcast` would be the quickest way I think. But Dev chains don't use that status, so in local setups, we will never see that.
`Ready` means that the basic checks have been done by the node (can pay fees, signed, and such). It means the xt has entered the tx pool

brenzi:
ok. let's merge this one

brenzi (approved):

PR #170, opened 2021-10-05
Fixed two JsonRPC request params

trevor-crypto:
Apparently, non-paritylabs JsonRPC endpoints (ex. [Elara](https://docs.elara.patract.io/01.JSON-RPC%20Methods--nc,d2/01%20Polkadot&Kusama.html#_5-9-state-getmetadata)) cannot handle
 the params field without it being in an array. The paritylabs rpc
 handles either way.


brenzi:
thanks!

brenzi (approved):

PR #174, opened 2021-11-03
remove scs node-template dependency

### LINKED ISSUE #165: Drop custom substrate-api-client-test-node pendency.
It seems that currently this is only prevented due to `SignedBlock` being exported in the node-template.

However, this is a simple type alias ,we could introduce in the api-client:

```
pub type SignedBlock = generic::SignedBlock<Block>;
```

clangenb:
Closes #165.

This was such a low-hanging fruit that I just needed to tackle it.

@haerdib can you update jenkins to use the node-template from substrate instead?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1395,7 +1395,7 @@ dependencies = [
 [[package]]
 name = "node-template-runtime"
 version = "3.0.0"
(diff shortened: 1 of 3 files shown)


clangenb:
@haerdib I think we can merge this one as is. I prefer not to invest work into jenkins anymore, and the tests pass there currently. We should rather tackle #176 soon.

haerdib (approved):
Approved with #176 :)

PR #177, opened 2021-11-08
Migrate whole CI to GithubActions

### LINKED ISSUE #176: move example tests from jenkins to github actions.
* #174 finally removes the need for the custom scs-node-template, we can test against a plain substrate node.

So we can simply find a substrate-node-template docker image and run the examples against it.

clangenb:
* Make GA actually fail upon clippy and fmt warnings
* Run all examples in GithubActions against a substrate node-template, freshly built from parity's repo. Closes #176 
* upload the build node-template as artifact.

Todo:
- [x] Merge and rebase: #181


haerdib (commented):
Looks good, thanks for cleaning that up.

Could you add some more description to the ci.yml? Like why you're excluding test-no-std from clippy and test?

Also: Does this PR close https://github.com/scs/substrate-api-client/issues/176?

clangenb:
Yes, agreed. I added some doc there. It does unfortunately not close #176. I haven't found a substrate node yet, which does not always update to latest.

clangenb:
Hopefully, I fixed Jenkins. I removed all the checks from Jenkins that we also run in GA.

Jenkinsfile
```diff
@@ -106,27 +57,8 @@ pipeline {
...
-              groovyScript(
-                parserId:'clippy-errors',
-                pattern: 'clippy_*.log',
-                reportEncoding: 'UTF-8'
-              )
           ]
         )
-        script {
-          try {
-            sh './ci/check_fmt_log.sh'
```
haerdib, line 123: I think with this change, the whole file `check_fmt_log.sh` is obsolete now. If so, can you delete it?

haerdib (approved):
Only one minor comment. Other than that it looks good

clangenb:
@brenzi two things:

* Moved all the integration tests to GithubActions and run them against the integritee-node. However, two examples don't run as they use the call enum from the node-template, which deviates from the integritee-node, is this acceptable?
* can you update the protected branch settings to watch the new statuses?

brenzi:
I removed status checks and I'll re-add them once we have merged this on and I have a clean setup.
Two failing examples isn't nice. I'd rather have a long-running parallel job that tests all the examples with a fresh build of node-template each time (but we should provide the node-template binary artifact and possibly include it in releases in that case, so we can adjust tutorials to simplify testing

clangenb:
I just found in subxt a random substrate latest node:

https://github.com/paritytech/subxt/blob/78fd590d17cdc21657c74837a066ea4d6e9beeaa/.github/workflows/rust.yml#L26

I added it here, but of course, the tests fail as it already uses metadata v14.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+        example: [ example_generic_extrinsic,
+                   example_print_metadata,
+                   example_transfer,
+                   example_get_storage,
+                   example_get_blocks,
+                   example_event_error_details,
+                   example_sudo,
+
+                    # those two fail if they are run against the integritee-node
+#                   example_benchmark_bulk_xt,
```
brenzi, line 70: these are important examples. And I would be in favor of using the real node-template for CI. A dependency on integritee-node will haunt us sooner or later

brenzi (changes requested):

haerdib (approved):
LGTM 👍

haerdib (approved):
Approving again

.github/workflows/ci.yml
clangenb: We now run all examples against the substrate-node-template from the master branch.

brenzi (approved):

PR #180, opened 2021-11-10
Extract node-api from the api-client main crate

clangenb:
Always better to keep stuff modular.


primitives/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+codec = { package = 'parity-scale-codec', version = "2.0.0", default-features = false,  features = ['derive']}
+
+sp-core = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+
+[features]
+default = ["std"]
+std = ["codec/std"]
```
haerdib, line 15: Could you add sp-core/std as well? Sorry to be picky here, but I've run into problems already due to missing std imports.

haerdib (approved):
Looks very clean, thanks. Only one small comment

primitives/Cargo.toml
clangenb: this is absolutely not picky, but a necessity! I know that these things can cause unforeseen problems!

PR #181, opened 2021-11-10
Cargo 2021 Edition + Create Workspace

clangenb:
Extracted all the Non-CI changes from #177 to make it cleaner.


haerdib (approved):
Thanks for separating these two, looks very clean now ! :)

PR #184, opened 2021-11-12
Add querying for payment info

pmikolajczyk41:
Extending API with `get_payment_info`. Thanks to that it is possible to know the `unadjusted_fee` for transaction, i.e. a raw fee before applying configurable multipliers. Useful for testing fee adaptation in e2e tests.


clangenb (approved):
Nice, seems useful! Thanks.

PR #185, opened 2021-11-13
rustfmt according to subxt

clangenb:
I am honestly doubting if we should merge that.

I have tried both, subxt and substrate fmts now.The substrate diff is huge, subxt much smaller. Both have changes that are not really my preferences. Shall we just preserve the plain formatting?

In the other repositories,  we regularly performed upstream merges, which is less frequently here. It would be for infrequent subxt merges only.


brenzi:
we've only merged with subxt a few times so far and things are getting more nd more stable, so we don't need to follow substrate nor subxt fmt rules necessarily IMO

haerdib:
In my opinion this rustfmt file is a mess. They have written down pretty much all values, be it default or not, and now it's impossible to see what has been changed and what not..

Regarding the changes: I don't mind if it's huge, as long as we agree that's the style we want to go with. But this is vastly different compared to all the other fmt's we're using, once again.

clangenb:
I agree that it looks like a mess, but it highlights every entry that is not the default. So we now know at least what the default values are. Regradless, I'd prefer a minimal rustfmt file.

clangenb:
Will close this, maybe we care about this later.

PR #186, opened 2021-11-19
Independent compose-macros crate

### LINKED ISSUE #84: make extrinsics an independent crate
I wanted to access the types we define in the extrinsics module on their own without accessing functionality of th e api-client itself multiple times already.

I suggest we make it a crate on its own to be able to import it independently.

clangenb:
It is in general best practice to make crates containing macros minimalistic, as they are not [hygienic](https://doc.rust-lang.org/reference/macros-by-example.html#hygiene).

Minimalistic macro crates reduce the risk of having unpredicted dependency resolution issues.

* Closes #84


primitives/src/extrinsics.rs
```diff
@@ -15,31 +15,32 @@
 
 */
 
-use sp_std::prelude::*;
+//! Primitives for substrate extrinsics.
```
haerdib, line 18: ❤️

src/extrinsic/balances.rs
```diff
@@ -15,24 +15,19 @@
 
 */
 
-use codec::Compact;
+//! Extrinsics for `pallet-balances`
```
haerdib, line 18:
```suggestion
//! Extrinsics for `pallet-balances`.
```

src/examples/example_benchmark_bulk_xt.rs
```diff
@@ -13,7 +13,7 @@
     limitations under the License.
 */
 
-//! This examples floates the node with a series of transactions
+//! This examples floats the node with a series of transactions
```
haerdib, line 16:
```suggestion
//! This examples floats the node with a series of transactions.
```

haerdib (approved):
Looks good, thanks for being hygienic :)

PR #197, opened 2021-12-14
Add instantiate_with_code

netrikare:
I added support for `instantiate_with_code` contracts API call + example usage.

### DIFF
--- a/src/examples/example_contract_instantiate_with_code.rs
+++ b/src/examples/example_contract_instantiate_with_code.rs
@@ -1,0 +1,101 @@
+/*
+    Copyright 2019 Supercomputing Systems AG
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


clangenb (commented):
Thank you very much for your contribution.

It looks good. However, could you make a remark in the example that it is community maintained and not CI tested; therefore it may not work as is? Could you also to the `extrinsic/contract` module that it is community maintained?

We won't integrate it in the CI, as we found out in the past that the effort is too big to maintain a custom node.

If you add these two things, I am happy to merge.

netrikare:
I added the message.

On a side note: I see you are testing against minimal `node-template` instead of `node`, which has all the default pallets built in. Maybe instead of building node-template in your CI you could simply pull `parity/substrate` docker image and test against it? No maintenance cost then.

clangenb:
I have only found a substrate docker node with a very old substrate version. All still v3.0.0, do you know a docker source with a recent substrate version? I would be very happy to use that.

netrikare:
Comparing commit hashes, it looks like `parity/substrate:latest` image tracks master branch.

clangenb:
Ok, cool. Thanks for the info. I will change that sometimes soon. 👍 

* #204

clangenb (approved):
Thx!

PR #198, opened 2021-12-22
bump substrate@b9cafb

clangenb:
Last commit before a breaking change where AccountId does no longer derive default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,15 +90,6 @@ dependencies = [
 
 [[package]]
 name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
-dependencies = [
- "winapi 0.3.9",
-]
-
-[[package]]
-name = "ansi_term"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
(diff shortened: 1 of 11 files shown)


haerdib (commented):
Happens if you increase the rust toolchain drastically 😄 

Regarding AccountId - before finding a substitution, such as encoding, we might consider switching to OptionQueries, these don't enforce default implementations. But I haven't looked into it yet regarding the substrate-api-client.

clangenb:
Yes, I think we will make that switch to the option queries. I also don't know if it requires changes to the api-client at all. I just wanted to give some context, why I did not pick the newest one.

Afaik however, this is only supported in the new frame v3 syntax. We don't have this syntax yet in any encointer-pallet...

Cargo.lock
```diff
@@ -3426,7 +3391,7 @@ version = "1.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1f559b464de2e2bdabcac6a210d12e9b5a5973c251e102c44c585c71d51bd78e"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if 0.1.10",
```
haerdib, line 3394: Unsure if that's a good change. Might be worth to rerun cargo update

node-api/src/error.rs
```diff
@@ -95,6 +95,9 @@ pub enum RuntimeError {
     /// At least one consumer is remaining so the account cannot be destroyed.
     #[error("At least one consumer is remaining so the account cannot be destroyed.")]
     ConsumerRemaining,
+    /// There are too many consumers so the account cannot be created.
```
haerdib, line 98: Seems to be quite redundant. Is it really necessary?

haerdib (approved):
LGTM. Minor comments, but feel free to merge if CI passes.

Cargo.lock
clangenb: This is very interesting, it is always this dependency, where this occurs. I have no idea why cargo does that.

node-api/src/error.rs
clangenb: I agree with you, I just took it from the original error in substrate.

PR #203, opened 2021-12-28
Move ApiError to separate file and add Send + Sync

haerdib:
Without Send + Sync it's not possible to use the ApiError within threads

(turned out to be a problem here : https://github.com/integritee-network/worker/blob/feature/fm-sidechain-peer-fetch-client/core/rpc-server/src/lib.rs#L68)

### DIFF
--- a/src/std/error.rs
+++ b/src/std/error.rs
@@ -1,0 +1,42 @@
+use crate::std::rpc::XtStatus;
+pub use ac_node_api::metadata::{InvalidMetadataError, MetadataError};
+
+pub type ApiResult<T> = Result<T, Error>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Fetching genesis hash failed. Are you connected to the correct endpoint?")]
+    Genesis,
+    #[error("Fetching runtime version failed. Are you connected to the correct endpoint?")]
+    RuntimeVersion,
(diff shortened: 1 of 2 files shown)


src/std/error.rs
```diff
@@ -0,0 +1,42 @@
+use crate::std::rpc::XtStatus;
+pub use ac_node_api::metadata::{InvalidMetadataError, MetadataError};
+
+pub type ApiResult<T> = Result<T, Error>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
haerdib, line 7: Everything was kept the same (simply copy pasted it) except for the Error `Other(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),` Where I added `+ Send + Sync + 'static`

src/std/mod.rs
```diff
@@ -1,7 +1,13 @@
+pub use crate::error::{ApiResult, Error as ApiClientError};
```
haerdib, line 1: public export, so no breaking changes should occur with this PR.

src/std/error.rs
clangenb: Ok, I am not 100% sure, if this is satisfied by most errors, but I guess you could just do

```
ApiClientError::Other(e.to_string().into())
```
So fine with me.

clangenb (approved):
Looks good to me.

PR #210, opened 2022-02-08
Remove unused default value in StorageMaps to allow OptionQuery

haerdib:
Removes unused default value from the StorageMaps. It was not used and blocked OptionQueries by throwing `MapValueTypeError`s when querying an Option Value.

Tested with recent node and worker (https://github.com/integritee-network/worker/pull/645)

### DIFF
--- a/node-api/src/metadata.rs
+++ b/node-api/src/metadata.rs
@@ -545,7 +545,7 @@ impl Metadata {
             .key())
     }
 
-    pub fn storage_map_key<K: Encode, V: Decode + Clone>(
+    pub fn storage_map_key<K: Encode>(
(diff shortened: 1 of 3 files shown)


clangenb (commented):
Looks so far good to me.

So, is there a difference now in how you call an `OptionQuery` map or an `ValueQuery`? If yes, it would be valuable to have that documented somewhere

haerdib:
No, there is no difference, actually. With the removal of the default value, it retrieves the optional storage value perfectly without any changes on the client side.

clangenb (approved):
Cool, then I approve as is.

PR #220, opened 2022-03-09
Bump substrate

### LINKED ISSUE #218: bump substrate (because of crate version increase in substrate master branch)
Substrate has again increased the version of various crates in their master branch, which implies that the api-client build is broken if we don't have the substrate crates in the cargo cache.

Bummer: We need to update the node-metadata crate because the metadata has changed

echevrier:
Closes #218.

Changes:
* bump substrate https://github.com/paritytech/substrate/commit/f5f286db0da99761792b69bf4a997535dcc8f260
* bump sp-runtime, sp-io, sp-keyring, sp-core, spc-rpc, sp-application-crypto to v6.0.0
* bump sp-version to v5.0.0
* bump scale-codec v3, bump scale-info v2.


clangenb:
Sorry, I did some minor beautification while debugging the broken no_std build: found this unfortunately: https://github.com/scs/substrate-api-client/issues/218#issuecomment-1063477099

Cargo.lock
```diff
@@ -1,3768 +0,0 @@
-# This file is automatically @generated by Cargo.
```
clangenb, line 1: Was this a conscious decision to remove the lock file? I am uncertain if I approve:

* Local builds are getting less reproducible
* We can no longer check in the lock file, which version works.

I know they say that libraries should not have a lock-file, but we have a hybrid case, we want to be able to build the example binaries.

clangenb, line 1: Was this a conscious decision to remove the lock file? I am uncertain if I approve:

* Local builds are getting less reproducible
* We can no longer check in the lock file, which version works.

I know they say that libraries should not have a lock-file, but we have a hybrid case, we want to be able to build the example binaries.

clangenb: I added it again for now, we could discuss that another time.

clangenb (approved):
Looks good to me now!

@haerdib Do you also want to have a quick look?

Cargo.lock
```diff
@@ -1810,10 +1837,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
+ "bitvec 0.20.4",
  "byte-slice-cast",
  "impl-trait-for-tuples",
- "parity-scale-codec-derive",
+ "parity-scale-codec-derive 2.3.1",
+ "serde",
+]
+
+[[package]]
+name = "parity-scale-codec"
```
haerdib, line 1848: We now have two versions of parity-scale-codec. Is this necessary?

I took a short look at the lock file, and it seems the old codec version is imported due to an old version of primitive-types, defined here: https://github.com/scs/substrate-api-client/blob/8a8f3059af1493e4eb9ccf114d98a4aea0dd1927/Cargo.toml#L21

haerdib (approved):
One small comment, otherwise it looks good :)

Cargo.lock
clangenb: I will fix this. Good catch.

PR #221, opened 2022-03-14
refactor: restructure examples

ivanceras:
This change just restructures the examples.
Putting example code in the examples/ folder automatically make it a runnable example.
There is also no need to make the files executable.


clangenb:
Cool, thank you very much! We will gladly merge this one after the integration-tests run. Interestingly, there is now a missing dependency for the examples, see the failing tests.

If you could add it to the deps again, the CI should work.

clangenb:
Unfortunately, there has been another issue popping up in the examples. I would be great if you could fix that.

PR #228, opened 2022-04-20
Bump substrate to polkadot-v0.9.19

echevrier:
Changes because of substrate:
- Add Transactional Error
- Module specific error value of ModuleError has changed

### DIFF
--- a/node-api/src/error.rs
+++ b/node-api/src/error.rs
@@ -121,7 +121,7 @@ impl RuntimeError {
                 error,
                 message: _,
             }) => {
-                let error = metadata.error(index, error)?;
+                let error = metadata.error(index, error[0])?;
                 Ok(Self::Module(PalletError {
                     pallet: error.pallet().to_string(),
                     error: error.error().to_string(),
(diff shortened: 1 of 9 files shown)


node-api/src/error.rs
```diff
@@ -121,7 +121,7 @@ impl RuntimeError {
                 error,
                 message: _,
             }) => {
-                let error = metadata.error(index, error)?;
+                let error = metadata.error(index, error[0])?;
                 Ok(Self::Module(PalletError {
```
echevrier, line 125: ModuleError has changed https://github.com/paritytech/substrate/blob/580ebae17fa30082604f1c9720f6f4a1cfe95b50/primitives/runtime/src/lib.rs#L476
But it seems that 0 value of error corresponds to the previous index

haerdib: For documentation purpose, relevant PR: https://github.com/paritytech/substrate/pull/10242

haerdib (approved):
LGTM. Only one uncertainity: As the substrate-api-client is used by others as well, should we merge the polkadot branch into the master? Or leave it as a branch?

@clangenb what's your opinion?

@echevrier can you rename the branhc (or create a new one, doesn't matter for me) to polkadot-v0.9.19 ? We should keep the naming consistent with substrate

PR #242, opened 2022-05-16
More cleanup in the examples to make it more lean

ivanceras:
- remove the `example_` prefix of all the examples
- hard code the url endpoint and remove the `cli.yml` and the code for getting the url in each of the examples to make it more lean
- remove deprecated code
- Remove the executable permission of some files that don't need to be executable: `Cargo.toml`, `README.md`, license.


haerdib:
@clangenb what do you think about this PR? I agree with the code smell, but that one can be fixed with a common function.

clangenb:
Hmmm, i agree with @ivanceras point that the examples should not be configurable. However, we did use the examples for some utility stuff with remote chains sometimes.

How about, we accept the PR and implement instead a simple 'substrate-api-cli'? This would probably be much more useful.

haerdib:
Created an issue https://github.com/scs/substrate-api-client/issues/308.

@ivanceras that's a green light for this PR. Sorry for the long wait. Do you have the time to get this PR up to date? If not, as it's been quite some time, I can take over.

haerdib:
Closing this in favour of #345

PR #244, opened 2022-05-17
[exrinsics] make generic over signed extra

clangenb:
This is to make the extrinsics truly generic over the signed extra, which can now be completely configured outside this crate if needed


echevrier (approved):
LGTM

PR #248, opened 2022-06-01
[extrinsicParams] Make `GenericExtra` generic over tip

clangenb:
With this change, I could successfully use the asset tips in the encointer-client without significant changes. 🥳


echevrier (approved):
LGTM

PR #249, opened 2022-06-02
Use substrate latest docker image to run node in CI

### LINKED ISSUE #204: Use `parity/substrate:latest` docker image in CI.
This will reduce the amount the CI runs by 25 minutes. It also seems to track the latest substrate master branch.

Instead of building the substrate node ourselves, we should just run the examples against the docker image.

This whole build-step can be removed: https://github.com/scs/substrate-api-client/blob/03251c4332c5b7fa4d952f772ba4469db1b824b7/.github/workflows/ci.yml#L55

And here we can use the docker image: https://github.com/scs/substrate-api-client/blob/03251c4332c5b7fa4d952f772ba4469db1b824b7/.github/workflows/ci.yml#L127

murerfel:
Use `parity/substrate:latest` to run a node for the examples. Remove the build step where we build our own node template.

Closes #204

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
 
       - name: Run latest node
         run: |
-          docker run -p 127.0.0.1:9944:9944 parity/substrate:latest --dev &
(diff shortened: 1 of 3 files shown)


haerdib:
Closing this in favour of #293

PR #250, opened 2022-06-03
Remove unnecessary calls to to_string() to improve performance.

Niederb:
The function showed up on some of my benchmarks therefore it is at least somewhat relevant to performance.
We could also change the API to use &str but that would be a breaking API change


clangenb (approved):
Nice!

clangenb:
@Niederb we have a fmt issue.

Niederb:
@clangenb Let's try again... fingers crossed

PR #252, opened 2022-06-13
Add Check_Non_Zero to additionalSigned

echevrier:


clangenb (approved):
Cool, looks good to me! As the tests run fine, this does not make a difference.

Hence, I assume that the `()` is simply being skipped in the scale-codec, which kind of makes sense.

PR #253, opened 2022-06-21
Make node-api `no_std` compatible

### LINKED ISSUE #190: make `node-api` `no_std` compatible
I have peeked (1.5h) into making the `node-api` stuff no_std compatible. It should be possible, but implies some rewriting of the code:

* The core change would be handling `PortableForm` differently. This is kind of hard now, as it behaves differently in `std` and `no_std`, hence the surrounding code, needs to be able to handle that.

However, strictly `no_std` is not possible either, as `sp-io`, which lies at the very bottom of the `substrate` deps, assumes wasm in `no_std`, which means it will not compile to `std`, without its `std` feature.

The nice thing is, if the `node-api` is `no_std` compatible, we can make many features no-std compatible of the api-client.

I reckon the overall effort to be a day for that (me doing it). It might be worth to try.

clangenb:
* Closes #190 

This went more smoothly than I thought. Only a few trade-offs were necessary:

* Can't derive `Error` for the errors in `node-api`.
* A few types don't derive debug in `no_std`.

Slight bummer:
* While we essentially can handle the metadata in `no_std` no. We have currently no way of easily passing/decoding it in the enclave because the necessary derives are not there upstream: https://github.com/paritytech/frame-metadata/issues/39


primitives/src/extrinsic_params.rs
```diff
@@ -198,7 +199,7 @@ where
     pub fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
         self.0.using_encoded(|payload| {
             if payload.len() > 256 {
-                f(&blake2_256(payload)[..])
+                f(&BlakeTwo256::hash(payload)[..])
```
clangenb, line 202: `no_std` compatibility exists only in this form.

node-api/Cargo.toml
```diff
@@ -6,20 +6,45 @@ edition = "2021"
...
-sp-core = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "master" }
-sp-runtime = { version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "master" }
+frame-support = { version = "4.0.0-dev",  default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+frame-system = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-core = { version = "6.0.0", default-features = false, features = ["full_crypto"], git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-std = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-runtime = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+
+# need to add this for `no_std`
+sp-application-crypto = { version = "6.0.0", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master", features = ["full_crypto"] }
```
haerdib, line 29: Is this alright to not add this to the std feature as std?

node-api/src/error.rs
```diff
@@ -23,92 +23,55 @@ use crate::{
...
+
+#[cfg(not(feature = "std"))]
+use alloc::string::{String, ToString};
+
+#[cfg(not(feature = "std"))]
+use sp_std::vec::Vec;
 
 /// Error enum.
-#[derive(Debug, Error)]
+#[derive(Debug, From)]
```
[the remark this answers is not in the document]
haerdib: Wholeheartedly agreed.

node-api/src/error.rs
```diff
@@ -23,92 +23,55 @@ use crate::{
...
+use derive_more::From;
 use sp_core::crypto::SecretStringError;
 use sp_runtime::{transaction_validity::TransactionValidityError, DispatchError, ModuleError};
-use thiserror::Error;
+
+#[cfg(not(feature = "std"))]
+use alloc::string::{String, ToString};
+
+#[cfg(not(feature = "std"))]
+use sp_std::vec::Vec;
```
haerdib, line 34: in `node-api/src/lib.rs` you're using the alloc vec. Would probably make sense to stay consistent.

node-api/src/metadata.rs
```diff
@@ -25,57 +25,59 @@ use frame_metadata::{
...
-use frame_support::serde::Serialize;
 use scale_info::{form::PortableForm, Type, Variant};
 use sp_core::storage::StorageKey;
-use std::{collections::HashMap, convert::TryFrom};
+
+#[cfg(feature = "std")]
+use serde::Serialize;
+
+// We use `BTreeMap` because we can't use `HashMap` in `no_std`.
+use sp_std::collections::btree_map::BTreeMap;
```
haerdib, line 35: Could also be imported by alloc directly.

node-api/src/metadata.rs
```diff
@@ -25,57 +25,59 @@ use frame_metadata::{
...
+
+#[cfg(not(feature = "std"))]
+use alloc::{
+    string::{String, ToString},
+    vec,
+    vec::Vec,
+};
 
 /// Metadata error.
-#[derive(Debug, thiserror::Error)]
```
haerdib, line 34: Maybe add `From` here as well?

node-api/src/error.rs
```diff
@@ -23,92 +23,55 @@ use crate::{
...
-
-impl From<String> for Error {
-    fn from(error: String) -> Self {
-        Error::Other(error)
-    }
-}
-
 /// Runtime error.
-#[derive(Clone, Debug, Eq, Error, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
haerdib, line 60: Not sure if you left it out deliberately, but mby add `From` here as well? ?

haerdib (approved):
LGTM, small comments only. Thanks !

PR #259, opened 2022-07-06
Metadata sgx compatible

echevrier:
- required for  https://github.com/integritee-network/worker/issues/113
- metadata can be coded and decoded. Use of integritee frame-metadata fork  : https://github.com/integritee-network/frame-metadata


clangenb (approved):
Looks good to me! Seems small, but it's a big step! :)

PR #260, opened 2022-07-07
Merge master commit  431e030...

echevrier:
Add functions:
- metadata sgx compatible
- make disable_target_static_assertions and optional feature


clangenb (approved):
Looks good to me!

Beware, you might need to patch the metadata in the worker, otherwise there might be an error revolving around duplicate `Metadata` definitions.

PR #262, opened 2022-07-14
Provide a connect method that does not spawn a thread

Niederb:
This is a proposal to reduce the number of threads that we spawn. As far as I can see this is a backward compatible change as it does not affect the public API or the externally visible blocking behavior. Please check my reasoning carefully.

In many cases we spawn the thread in `start_rpc_client_thread(..)` and then immediately call the blocking `recv()` function. Then we have two threads that are waiting in a blocking manner. 
Instead we can make the `connect()` in a blocking manner directly and wait there. The `recv()` afterwards will then not be blocking anymore (as there is already data in the channel). This saves spawning a thread and additional scheduling overhead. 

See attached system call numbers to see the improvements when running our benchmark for 3 minutes. 
- Overhead due to system calls is down roughly 25%
- Number of threads is almost cut in half (calls to `clone()`)
  - Connect is creating another thread internally so there are still plenty of threads

[systemcalls-improved.txt](https://github.com/scs/substrate-api-client/files/9109065/systemcalls-improved.txt)
[systemcalls-baseline.txt](https://github.com/scs/substrate-api-client/files/9109066/systemcalls-baseline.txt)

### DIFF
--- a/src/std/rpc/ws_client/client.rs
+++ b/src/std/rpc/ws_client/client.rs
@@ -37,8 +37,7 @@ impl WsRpcClient {
 impl RpcClientTrait for WsRpcClient {
     fn get_request(&self, jsonreq: Value) -> ApiResult<String> {
         let (result_in, result_out) = channel();
-        self.get(jsonreq.to_string(), result_in)?;
-
+        self.direct_rpc_connect(jsonreq.to_string(), result_in, on_get_request_msg)?;
         let str = result_out.recv()?;
         Ok(str)
     }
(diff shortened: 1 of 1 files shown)


src/std/rpc/ws_client/client.rs
```diff
@@ -167,4 +165,20 @@ impl WsRpcClient {
...
+    ) -> WsResult<()> {
+        connect(self.url.as_str(), |out| RpcClient {
+            out,
+            request: jsonreq.clone(),
+            result: result_in.clone(),
+            on_message_fn,
+        })?;
+
+        Ok(())
+    }
```
clangenb, line 183: This looks good, I agree with your thinking, as the whole thing is blocking anyhow, this is merely an internal improvement. However, I wonder about one things:

* Would it not make sense to create the channel internally here, and directly return a `WsResult<String>`? Otherwise, the function looks the same from the outside as the `start_rpc_client_thread` even though it has different semantics.

clangenb (commented):
Very nice, I  have only one question!

clangenb (approved):
Nice, I like it a lot. I am very impressed how much this improves the benchmark!

PR #270, opened 2022-08-18
Sp/changes polkadot update

masapr:
do not merge, so it doesn't interfer with the current polkadot update

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -335,6 +335,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bytes"
+version = "1.2.1"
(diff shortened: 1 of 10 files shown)


haerdib (commented):
LGTM. You should probably run a downgrade of substrate.

polkadot-v.0.9.27 was on commit `5677b91eb6ad8e280cf7dcbdb8990d31bd2d74b1`

E.g. run
`cargo update -p sp-std --precise 5677b91eb6ad8e280cf7dcbdb8990d31bd2d74b1`

And check with @Niederb  :)

PR #276, opened 2022-09-09
Remove outdated tutorial

### LINKED ISSUE #275: substrate-test-nodes project used by tutorial is deprecated
substrate-test-nodes project used by tutorial is deprecated.  it also fails to build on macOS w/M2 cpu because of the ring v0.14.6 dependency.   the tutorial needs to be updated to show how to use the api-client perhaps with the standard substrate node template.

clangenb:
The `substrate-test-node` repository has been archived and does no longer compile. Hence, there isn't any node that the tutorial can be tested against. As our examples are pretty instructive too, I believe that we no longer need the tutorial.

* Closes #275


brenzi:
Do we really want to delete them? They sgould stilk work agaibst a substrate-template node, right?

clangenb:
It is only the substrate kitty tutorial that is gone. And no, it does not work woth the template node.

brenzi (approved):
Ok

PR #282, opened 2022-09-27
Polkadot v0.9.29

clangenb:
patch `ed25519-zebra` for no_std compatibility. See: https://github.com/paritytech/substrate/issues/12225


haerdib (commented):
LGTM! 

One remark: unlike all the other repositories, we keep the api-client master pointing to master branches of substrate. That's because multiple, independent projects are using this client.

So keep this branch as it is (polkadot-v0.9.29 should point to the equivalent substrate branch), but create a new PR with the same changes but with master branches. With
`cargo update -p sp-core --precise COMMITHASH you should be able to choose a commit with the same level as the polkadot v0.9.29  branch)

OverOrion:
Thank you @haerdib, opened the _to-be-merged_ PR: https://github.com/scs/substrate-api-client/pull/289.

Issue #274, opened 2022-09-06
Licensing to apache-2.0


Niederb:
At the moment there are some GPL licensed files in `node-api`. They were taken from https://github.com/paritytech/subxt/.
However subxt recently changed their licenses to dual GPL and Apache. See: https://github.com/paritytech/subxt/commit/c55b7ba3e34118f61dc51499a0b38262c513aec4
I think we should "re-import" the files and then we could license the `node-api` crate under Apache-2.0.
I think all the code in this repo is then Apache-2.0. I think we should also make the license for the other crates more explicit with `license = "Apache-2.0"` in the toml.

PR #283, opened 2022-09-29
Set license explicitly to Apache-2.0 and set SCS as author

Niederb:
Set the license in all `Cargo.toml` files except for `node-api` (see #274 ). This will improve the license check in our other repos downstream (worker, ...)


clangenb (commented):
Nice. LGTM.

clangenb (approved):

PR #285, opened 2022-09-29
Fix substrate node build in CI

clangenb:


haerdib (approved):
Nice one !

Niederb:
I guess the `uses: arduino/setup-protoc@v1` was what fixed the build? Do you now why this is suddenly needed?

clangenb:
I have no idea, but it is used to generate rust code based on some non-rust input files. Maybe they added this recently (I have not checked).

PR #290, opened 2022-10-21
GHA: update `rust-cache` to v2, as v1 contains deprecated commands (`save-state` and `set-output`)

OverOrion:
>Starting today runner version 2.298.2 will begin to warn you if you use the save-state or set-output commands via stdout. We are monitoring telemetry for the usage of these commands and plan to fully disable them on 31st May 2023. Starting 1st June 2023 workflows using save-state or set-output commands via stdout will fail with an error.
---

See https://github.com/Swatinem/rust-cache/issues/80 and https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       - name: init-rust-target
         run: rustup show
 
-      - uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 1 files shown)


clangenb (approved):
Nice, always good to have this maintenance!

PR #291, opened 2022-10-31
Add example of using account seed

haerdib:
rebased work of #216


clangenb (commented):
I think this example is a bit an overkill and very redundant with the other one. shall we not just use the seed approach in the other example?

We could also add a comment, that you could use the keyring alternatively, like:

```rust 

// Alice's seed: subkey inspect //Alice
//
// Alternatively use: let from = Account::Keyring::Alice.pair();
let from: sr25519::Pair = Pair::from_string(
"0xe5be9a5092b81bca64be81d212e7f2f9eba183bb7a90954f7b76361f6edb5c0a",
None,
)
.unwrap();
```

haerdib:
Agreed, let's do that

PR #293, opened 2022-11-01
Use substrate kitchensink-runtime docker image in CI

### LINKED ISSUE #204: Use `parity/substrate:latest` docker image in CI.

haerdib:
Use `parity/substrate:latest` to run a node for the examples. Removes the need to build our own node template.
- The substrate docker image does not use the node-template runtime but the [`kitchensink-runtime`](https://github.com/paritytech/substrate/blob/c11b2621e6e0e20263a230b783d68820370f9ee1/bin/node/runtime/src/lib.rs). 
Good thing: it now includes more pallets, leaving more options to test examples in CI.
- update to substrate commit `c24431eb6a95197550b30715a4c124be1aea79de` to include a substrate fix for the uniques  pallet: https://github.com/paritytech/substrate/pull/12297

Closes #204

Replaces PR #249

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,14 +69,6 @@ jobs:
 
   examples:
     runs-on: ${{ matrix.os }}
-
-#    services:
-#      latest_node:
-#        image: parity/substrate:latest
-#        ports:
-#        - 9944:9944
-#        options: --dev --ws-external
-
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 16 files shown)


murerfel (approved):
I cannot really judge or review the substrate specific parts, but the build setup looks fine to me 👍

PR #294, opened 2022-11-01
Added features for staking and payout 

shanithkk:
# Description

Added staking functionalities and payout example. fixed bugs in staking modules

Fixes # Staking modules

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Added test case in example folder.

- [x] example_staking_payout

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules


node-api/src/error.rs
```diff
@@ -103,6 +103,7 @@ impl RuntimeError {
                 Ok(Self::Other("transactional error".into()))
             }
             DispatchError::Other(msg) => Ok(Self::Other(msg.to_string())),
+            _ => todo!(),
```
haerdib, line 106: This is only necessary because the substrate version has been updated (due to the new dependency). I'd rather have this update in a separate PR if possible. You can downgrade the substrate version easily with the command
`cargo update -p sp-std --precise f1c60e529a6c1d158a0a37fe0a8daf887b2b9f74`.

src/extrinsic/common.rs
```diff
@@ -0,0 +1,40 @@
...
+pub struct BatchPayout {
+    pub calls: Vec<([u8; 2], PayoutStakers)>,
+}
+
+#[derive(Clone, Eq, PartialEq, Encode, Decode, Debug)]
+pub struct ForceEra {}
+
+#[derive(Clone, Eq, PartialEq, Encode, Decode, Debug)]
+pub struct Batch {
+    pub calls: Vec<RuntimeCall>,
```
haerdib, line 39: A common import from the template runtime should not be necessary. The Api Client should be runtime independent, respectively, work for all runtimes.

Can you change this to a more generic implementation? E.g. 
```rust
pub struct Batch<Call> {
    pub calls: Vec<Call>,
}
```
?

src/extrinsic/common.rs
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+// Common types.
+
+use codec::{Decode, Encode};
+use node_template_runtime::RuntimeCall;
+use sp_runtime::AccountId32;
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub struct PayoutStakers {
+    pub validator_stash: AccountId32,
```
haerdib, line 25: Can you change this to a `GenericAddress`?

src/extrinsic/utility.rs
```diff
@@ -0,0 +1,63 @@
...
+{
+    pub fn batch_payout_stakers(
+        &self,
+        calls: Vec<([u8; 2], PayoutStakers)>,
+    ) -> UtilityBatchPayoutXt<Params::SignedExtra> {
+        let calls = BatchPayout { calls };
+        compose_extrinsic!(self, UTILITY_MODULE, UTILITY_BATCH, calls)
+    }
+
+    pub fn batch(&self, calls: Vec<RuntimeCall>) -> UtilityBatchXt<Params::SignedExtra> {
```
haerdib, line 54: We shouldn't depend on a specific runtime for composing the call. Can you remove this dependency?

examples/example_staking_payout.rs
```diff
@@ -0,0 +1,91 @@
...
+    let account: AccountId;
+    match AccountId32::from_ss58check("5DJcEbkNxsnNwHGrseg7cgbfUG8eiKzpuZqgSph5HqHrjgf6") {
+        Ok(address) => account = address,
+        Err(e) => panic!("Invalid Account id : {:?}", e),
+    }
+    // initialize api and set the signer (sender) that is used to sign the extrinsics
+    let pair = sr25519::Pair::from_string(/* Give Your Secret Key for test, dont upload in repo */"", None).unwrap();
+    let api = api.set_signer(pair.clone());
+
+    let active_era: Staking = api
```
haerdib, line 42: Instead of defining your own `Staking` struct, which will get easily outdated, you could take it directly from the pallet: https://github.com/paritytech/substrate/blob/master/frame/staking/src/lib.rs#L352-L361

examples/example_staking_payout.rs
```diff
@@ -0,0 +1,91 @@
...
+        .get_storage_value("Staking", "ActiveEra", None)
+        .unwrap()
+        .unwrap();
+    println!("{:?}", active_era);
+    let idx = active_era.index - 1;
+    match api
+        .get_storage_double_map("Staking", "ErasStakers", idx, &account, None)
+        .unwrap()
```
haerdib, line 50: I think simply adding an additional unwrap() instead of the match statement would make the code more readable is sufficient for an example.

examples/example_staking_payout.rs
```diff
@@ -0,0 +1,91 @@
...
+        let call = api.payout_stakers(idx, account.clone());
+        let result = api
+            .send_extrinsic(call.hex_encode(), XtStatus::InBlock)
+            .unwrap();
+        println!("{:?}", result);
+    }
+}
+
+pub fn get_node_url_from_cli() -> String {
+    let url = format!("wss://westend-rpc.polkadot.io:443"); //wss://rpc.polkadot.io:443
```
haerdib, line 67: Why not take the url from the cli, like the other examples?

examples/example_staking_payout.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug)]
+pub struct Staking {
+    index: u32,
+    start: Option<u32>,
+}
+
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug, Default)]
+pub struct Exposure<_0, _1> {
```
haerdib, line 79: Again, take this directly from the substrate pallet.

examples/example_staking_payout.rs
```diff
@@ -0,0 +1,91 @@
...
+    #[codec(compact)]
+    pub total: _1,
+    #[codec(compact)]
+    pub own: _1,
+    pub others: Vec<IndividualExposure<_0, _1>>,
+}
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, Debug, Default)]
+pub struct IndividualExposure<_0, _1> {
```
haerdib, line 87: ^^^^

src/extrinsic/staking.rs
```diff
@@ -49,31 +60,49 @@ pub type StakingWithdrawUnbondedFn = (CallIndex, u32);
...
+    UncheckedExtrinsicV4<StakingForceNewEraFn, SignedExtra>;
+pub type StakingForceNewEraAlwaysXt<SignedExtra> =
+    UncheckedExtrinsicV4<StakingForceNewEraAlwaysFn, SignedExtra>;
+pub type StakingForceNoEraXt<SignedExtra> = UncheckedExtrinsicV4<StakingForceNoEraFn, SignedExtra>;
+pub type StakingSetPayeeXt<SignedExtra> = UncheckedExtrinsicV4<StakingSetPayeeFn, SignedExtra>;
+pub type StakingSetValidatorCountXt<SignedExtra> =UncheckedExtrinsicV4<StakingSetValidatorCountFn, SignedExtra>;
 
 // https://polkadot.js.org/docs/substrate/extrinsics#staking
-impl<P, Client> Api<P, Client>
+impl<P, Client, Params> Api<P, Client, Params>
```
haerdib, line 91: To me it looks like you only need the `SignedExtra`, not the full `Params` trait. Is there a reason you're taking the  Params as input?

haerdib (commented):
Thanks a lot for your contribution! Some nitpicking comments, but awesome work :)

FYI: 
 I'm currently working on a PR that might change the CI from using the template node to the substrate `kitchensink-runtime` (https://github.com/paritytech/substrate/blob/master/bin/node/runtime/Cargo.toml), which includes the Staking pallet. It would be awesome if we can include your example in CI then.

src/extrinsic/common.rs
shanithkk: Actually its throwing an error like  "value: InvalidHexString(InvalidStringLength)'" after changing to GenericAddress type.

haerdib: Okay. Let's leave as is for now, then. Thanks for trying!

node-api/src/error.rs
shanithkk: Resolved

src/extrinsic/staking.rs
shanithkk: I couldn't make it as only signedExtra because of it is a type inside the trait, any suggestion?

examples/example_staking_payout.rs
shanithkk: changed as per your suggestion

src/extrinsic/staking.rs
haerdib: Leave as is for now. It's fine. Thanks for looking into it 👍

haerdib (approved):
Awesome. Thanks a lot!

PR #296, opened 2022-11-02
Refactor ws_client to remove unwraps and duplicate code

### LINKED ISSUE #241: More robust error handling
Receiving some unexpected error/msg (e.g. a json message without the `error` field) makes the calling thread panic, [src/std/rpc/ws_client/mod.rs](https://github.com/scs/substrate-api-client/blob/ab586d8be719243aa29503ec8c7b1457b2cd0830/src/std/rpc/ws_client/mod.rs#L342). Instead of panicing, please consider handle it some other way, e.g. return some `unknown error`.

haerdib:
- Removes unused `XtStatus::Error`
- Removes potential panics inside the ws_client
- Refactor ws_client message handling to expect a struct `MessageHandler` implementing a trait `HandleMessage` instead of a function input. Removes the need of duplicating a lot of code.
- Adds values to `XtStatus` to be able to differentiate statuses according to value.

fixes #241

### DIFF
--- a/src/std/rpc/mod.rs
+++ b/src/std/rpc/mod.rs
@@ -32,6 +32,8 @@ pub enum RpcClientError {
 	Extrinsic(String),
 	#[error("mpsc send Error: {0}")]
 	Send(#[from] std::sync::mpsc::SendError<String>),
+	#[error("Expected some error information, but nothing was found")]
+	NoErrorInformationFound,
 }
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
--- a/src/std/rpc/mod.rs
+++ b/src/std/rpc/mod.rs
@@ -38,14 +38,14 @@ pub enum RpcClientError {
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum XtStatus {
-	Future = 10,
-	Finalized = 5,
-	InBlock = 4,
-	Broadcast = 3,
-	Ready = 2,
+	Unknown = 0,
 	/// uses `author_submit` without watching.
 	SubmitOnly = 1,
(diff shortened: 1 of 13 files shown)


src/std/rpc/ws_client/mod.rs
```diff
@@ -141,149 +143,113 @@ where
 	}
 }
 
-pub fn on_get_request_msg(msg: Message, out: Sender, result: ThreadOut<String>) -> WsResult<()> {
```
haerdib, line 144: The diff from here on is a little messed up.
The `on_get_request_msg` has not changed (only moved to be implemented by `GetRequestHandler`, same goes for the `on_subscription_msg`). The main change: `SubmitAndWatchHandler` now substitutes all the `on_extrinsic_msg_until_*` functions.

src/std/rpc/mod.rs
```diff
@@ -32,20 +32,20 @@ pub enum RpcClientError {
...
-	Error,
-	Unknown,
+	Future = 10,
+	Finalized = 5,
+	InBlock = 4,
+	Broadcast = 3,
+	Ready = 2,
+	/// uses `author_submit` without watching.
+	SubmitOnly = 1,
+	Unknown = 0,
```
clangenb, line 48: I would vote to sort this according to the numbering. :)

clangenb (commented):
Only small comments, looks good otherwise!

src/std/rpc/mod.rs
haerdib: Done 🍡

clangenb (commented):
Thanks, I like this PR!

clangenb (approved):

PR #312, opened 2022-11-11
Future status returns Future error instead of HexError

### LINKED ISSUE #157: Upon "Future" status, api-client returns `Err(InvalidHexString)`
If an extrinsic is placed in the future pool, the api-client will issue a [warning](https://github.com/scs/substrate-api-client/blob/2edd060c0a515718d62f3a78008b4123a5acefeb/src/std/rpc/ws_client/mod.rs#L206) and then end the process with a [None value](https://github.com/scs/substrate-api-client/blob/2edd060c0a515718d62f3a78008b4123a5acefeb/src/std/rpc/ws_client/mod.rs#L207).

The None is turned into an [empty String](https://github.com/scs/substrate-api-client/blob/2edd060c0a515718d62f3a78008b4123a5acefeb/src/std/rpc/ws_client/mod.rs#L293) and sent back [via channel](https://github.com/scs/substrate-api-client/blob/2edd060c0a515718d62f3a78008b4123a5acefeb/src/std/rpc/ws_client/mod.rs#L299)

If one is waiting until "Finalized", the api client [expects an hex string](https://github.com/scs/substrate-api-client/blob/2edd060c0a515718d62f3a78008b4123a5acefeb/src/std/rpc/ws_client/client.rs#L64). However, because the return value is empty it will finally return an hex error `Err(InvalidHexString(InvalidStringLength))`.

```bash
[2021-09-01T14:29:02Z DEBUG substrate_api_client::rpc::client] got msg {"jsonrpc":"2.0","method":"author_extrinsicUpdate","params":{"result":"future","subscription":"18PvlhIHlZgUWISr"}}
[2021-09-01T14:29:02Z WARN  substrate_api_client::rpc::client] extrinsic has 'future' status. aborting
[2021-09-01T14:29:02Z DEBUG substrate_api_client::rpc::client] Thread end result :Sender { .. } value:None
[2021-09-01T14:29:02Z INFO  substrate_api_client] finalized:
[2021-09-01T14:29:02Z ERROR substratee_worker::polkadex_db] [<] Extrinsic got finalized. Hash: Err(InvalidHexString(InvalidStringLength))
```

Should the client wait until future has been finalized? Or should we change the return value handling?

### LINKED ISSUE #202: support custom nonce when transfer tx

haerdib:
- Adds a custom nonce example which additionally checks the future return error
- The `ThreadOut Message` is now more generic and is defined by the struct implementing the `HandleMessage` struct.
- ⚡ `GetRequestHandler`, `SubmitOnlyHandler` and `SubmitAndWatchHandler`, e.g. all `WsRpcClient` functions except for the `start_subscriber` are now returning an `Result<Option<String>>` instead of just a `String`. This way, a RpcClientError can be forwarded to the receiving thread. Additionally, a `None` is now allowed, which is quite commonly used in substrate. Most notable result: `send_extrinsic` can now forward the received result and does not depend on the "expected" xt_status any more (making it effectively unable to handle unexpected results, such as Future status).
- `Error::RpcClient()` now expects an `RpcClientError`, not a `String` any more. To not lose any flexibility regarding user-implementation, `RpcClientError` now implements an `Other(#[from] Box<dyn std::error::Error + Send + Sync + 'static>)`.

example_custom_nonce now prints the following error:

```Retrieved error RpcClient(Extrinsic("Unexpected extrinsic status: Future, stopped watch process prematurely."))```

closes #157
closes #202 (added a specific custom nonce example)

### DIFF
--- a/src/std/rpc/ws_client/client.rs
+++ b/src/std/rpc/ws_client/client.rs
@@ -15,7 +15,7 @@
 
 */
 
-use super::{HandleMessage, ThreadMessage};
+use super::HandleMessage;
 use crate::{
 	rpc::ws_client::{
 		GetRequestHandler, SubmitAndWatchHandler, SubmitOnlyHandler, SubscriptionHandler,
@@ -32,6 +32,7 @@ use log::info;
 use serde_json::Value;
 use sp_core::H256 as Hash;
 use std::{
+	fmt::Debug,
 	sync::mpsc::{channel, Sender as ThreadOut},
 	thread,
 };
@@ -81,7 +82,7 @@ impl Subscriber for WsRpcClient {
 	fn start_subscriber(
 		&self,
 		json_req: String,
-		result_in: ThreadOut<ThreadMessage>,
+		result_in: ThreadOut<<SubscriptionHandler as HandleMessage>::ThreadMessage>,
 	) -> Result<(), ws::Error> {
 		self.start_subscriber(json_req, result_in)
 	}
@@ -89,14 +90,18 @@ impl Subscriber for WsRpcClient {
 
 #[allow(clippy::result_large_err)]
 impl WsRpcClient {
-	pub fn get(&self, json_req: String, result_in: ThreadOut<ThreadMessage>) -> WsResult<()> {
+	pub fn get(
+		&self,
+		json_req: String,
+		result_in: ThreadOut<<GetRequestHandler as HandleMessage>::ThreadMessage>,
+	) -> WsResult<()> {
 		self.start_rpc_client_thread(json_req, result_in, GetRequestHandler::default())
 	}
 
 	pub fn send_extrinsic(
 		&self,
 		json_req: String,
-		result_in: ThreadOut<ThreadMessage>,
+		result_in: ThreadOut<<SubmitOnlyHandler as HandleMessage>::ThreadMessage>,
 	) -> WsResult<()> {
 		self.start_rpc_client_thread(json_req, result_in, SubmitOnlyHandler::default())
 	}
@@ -104,7 +109,7 @@ impl WsRpcClient {
 	pub fn send_extrinsic_until_ready(
 		&self,
 		json_req: String,
-		result_in: ThreadOut<ThreadMessage>,
+		result_in: ThreadOut<<SubmitAndWatchHandler as HandleMessage>::ThreadMessage>,
 	) -> WsResult<()> {
 		self.start_rpc_client_thread(
 			json_req,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/std/rpc/ws_client/mod.rs [hunk 1]
```diff
@@ -239,38 +273,49 @@ impl SubmitAndWatchHandler {
...
 		match parse_status(return_string) {
 			Ok((xt_status, val)) => {
 				if xt_status as u32 >= 10 {
-					error!("Node returned unexpected {:?}, ending watch process.", xt_status);
-					end_process(out, result, val)?;
+					let error = RpcClientError::Extrinsic(format!(
+						"Unexpected extrinsic status: {:?}, stopped watch process prematurely.",
+						xt_status
+					));
+					end_process(out, result, Err(error))?;
```
haerdib, line 293: The error is now forwarded to the receiving thread

src/std/rpc/ws_client/mod.rs
```diff
@@ -155,22 +169,38 @@ fn extrinsic_has_failed(event_details: &EventDetails) -> bool {
...
+		result
+			.send(result_str)
+			.map_err(|e| Box::new(RpcClientError::Send(format!("{:?}", e))).into())
 	}
 }
 #[derive(Default, Debug, PartialEq, Eq, Clone)]
 pub struct SubscriptionHandler {}
 impl HandleMessage for SubscriptionHandler {
-	fn handle_message(&self, msg: Message, out: Sender, result: ThreadOut<String>) -> WsResult<()> {
+	type ThreadMessage = String;
```
haerdib, line 196: Subscriptions still return a String only because subscription is an ongoing thing and multiple answers are expected. 
Therefore, it should be fine to handle Errors and None values internally and to simply log them.

src/std/rpc/ws_client/client.rs
```diff
@@ -65,12 +68,12 @@ impl RpcClientTrait for WsRpcClient {
...
 		};
 
-		let response = self.direct_rpc_request(jsonreq, SubmitAndWatchHandler::new(exit_on))?;
-		info!("Got response {} from waiting for {:?}", response, exit_on);
-		match exit_on {
-			XtStatus::Finalized | XtStatus::InBlock => Ok(Some(Hash::from_hex(response)?)),
-			XtStatus::Broadcast | XtStatus::Ready | XtStatus::SubmitOnly => Ok(None),
-			_ => Err(ApiClientError::UnsupportedXtStatus(exit_on)),
+		let maybe_response =
+			self.direct_rpc_request(jsonreq, SubmitAndWatchHandler::new(exit_on))??;
```
haerdib, line 72: Two `??` are necessary because two types of error may happen:
1.  ws connection error or channel error in the function `direct_rpc_request`
2. An Error of the `ThreadOut<ThreadMessage>`, e.g. an `RpcClientError`.

clangenb (approved):
Nice, looks good to me!

PR #320, opened 2022-11-15
abstract api client

haerdib:
Not compiling, draft PR only. But it's little more work than anticipated


haerdib:
Closing this, has been replaced by several PRs.

PR #321, opened 2022-11-16
Add timeout to rpc client and fn wait_for_event

### LINKED ISSUE #134: api-client never returns if node does not answer
In case the node does not answer (or answers in an unexpted manner, due to whatever reason), the api-client will end up in a never ending loop here: https://github.com/scs/substrate-api-client/blob/master/src/lib.rs#L501 

A solution would be to exchange `.recv()` with `recv_timout()`. To provide maximum flexibility, either an extra function input parameter would be necessary (for example `Option<Duration>` [here](https://github.com/scs/substrate-api-client/blob/master/src/lib.rs#L491)) or an additional struct parameter could be introduced in the future [WsRpcClient](https://github.com/trevor-crypto/substrate-api-client/blob/rpcclient/src/std/rpc/ws_client/client.rs#L25)

haerdib:
- adds a timeout to the `WsRpcClient`. By default it is 300 secs ( = 5 minutes). Finalization may take quite some time, so I think that's a reasonable amount of time.
- the `wait_for_event` functions now take an additional optional timeout parameter. Because the event may be expected to happen somewhen far in the future, I did not want to introduce a default timeout just like that. 

closes #134

### DIFF
--- a/examples/example_benchmark_bulk_xt.rs
+++ b/examples/example_benchmark_bulk_xt.rs
@@ -19,10 +19,8 @@
 // > substrate-test-node --dev --execution native --ws-port 9979 -ltxpool=debug
 
 use clap::{load_yaml, App};
-
 use kitchensink_runtime::{BalancesCall, RuntimeCall};
 use sp_keyring::AccountKeyring;
-
 use substrate_api_client::{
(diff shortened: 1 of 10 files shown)


haerdib:
Closig this PR - doesn't make sense currently because `connect()` from `ws` is blocking. So `recv_timeout` does not help. As the ws server will be exchanged anyway I think, I'll close for now.

PR #332, opened 2022-11-18
Bump version of all crates

haerdib:
Reasoning: The licence header has been changed since v0.1.0. I think that's reason enough for bumping everything


clangenb (approved):
I agree, lgtm!

Issue #267, opened 2022-08-01
Abstract api-client main features behind a trait


clangenb:
This would greatly simplify the separation between the core logic and the custom business logic in the integritee-worker in case the node has some customizations, which are incompatible with the default api-client type, e.g., the signed extra if  `pallet_asset_tx_payment` is used.

Excerpt from the envisioned trait

```rust

pub trait Api {
          fn get_storage<V>() -> Result<Option<V>>;
     
          fn subscribe_events();

        // etc.
}
```

PR #335, opened 2022-11-21
Restructure std folder

haerdib:
Preparatory step for #267 

Refactor only, no logical changes.

The following was changed:
- `Api` implementation has been moved to `api.rs`
- `XtStatus` and `ReadProof` have been moved to one level above, because they are not rpc specific
- Moved subscription implementation to separate file and to the api level.

Goal of this PR: If the ws-client is exchanged, only the rpc folder needs to be updated.


src/std/rpc/ws_client/mod.rs
```diff
@@ -15,35 +14,37 @@ use ac_node_api::Events;
...
 
+#[allow(clippy::result_large_err)]
+pub trait Subscriber {
+	fn start_subscriber(
+		&self,
+		json_req: String,
+		result_in: ThreadOut<String>,
+	) -> Result<(), WsError>;
+}
+
```
clangenb, line 47: Why didn't you move this one to the subscription module?

clangenb (commented):
Only one question, looks good otherwise:-)

src/std/rpc/ws_client/mod.rs
haerdib: Good question. Original thought: Because the api does not implement the subscription trait. But that means, the api is still in the wrong place.. Should not be in the ws_client folder. Let me think on that.

haerdib: Better now?

Added a "goal" to the PR description, which hopefully makes it clear what I want out of this PR.

clangenb (approved):
Looks good to me now!

echevrier (approved):
LGTM

PR #336, opened 2022-11-22
Decouple RpcClient and Api

haerdib:
While RpcClient and the Api implementation have been somewhat decoupled, they were still quite coupled (E.g. `RpcClient Trait` returned an `ApiResult`).
This PR aims to decouple them more by doing the following:
- move rpc implementation to a separate folder. Reasoning: `extrinsic` as well as `api` depend on the `RpcClient` trait.
- renamed the std folder to api. Reasoning: Extrinsic was `std` only as well, so std did not make sense anyway. It's fairly easy visible in the `lib.rs` file which features are available in std and which are not.
- `RpcClient` trait now return as `RpcError`.


clangenb (approved):
I approve the `std` folder has been bothering me too.

PR #337, opened 2022-11-22
[Api] add `update_runtime` function and return metadata from cache

haerdib:
💥  **Breaking Changes**:  
- `Api` variables are no longer public. They should be accessed via the getter methods
- `get_metadata`, `get_spec_version`, `get_genesis_hash` are now private. The user should access these values via the cached values in the `Api` struct.
    ❗  Careful: In case one relied on `get_metadata` returning the most recent metadata from the node, that's not the case any more. If that's necessary, run `update_runtime()` before calling the cache via `metadata()`.
- Setter methods (such as `set_signer`) now take a mutable reference and do not return a `Self` anymore.
 

💚  **Feature Changes**: 
- Added `update_runtime` function to update the Api in case of a runtime upgrade
- removes extra `_get_request` function: RpcClient now directly returns an `Option` without mapping it to an empty string and back to an Option.
- Added some descriptive comments and split up the `impl Api` for a more sensible layout.

### DIFF
--- a/node-api/src/metadata.rs
+++ b/node-api/src/metadata.rs
@@ -17,10 +17,8 @@ use frame_metadata::{
 	StorageEntryMetadata, META_RESERVED,
 };
 use scale_info::{form::PortableForm, PortableRegistry, Type};
-use sp_core::storage::StorageKey;
-
-#[cfg(feature = "std")]
 use serde::Serialize;
+use sp_core::storage::StorageKey;
 
 // We use `BTreeMap` because we can't use `HashMap` in `no_std`.
 use sp_std::collections::btree_map::BTreeMap;
--- a/node-api/src/metadata.rs
+++ b/node-api/src/metadata.rs
@@ -67,7 +67,7 @@ pub enum MetadataError {
 	IncompatibleMetadata,
 }
 
-#[derive(Clone, Debug, Encode, Decode)]
+#[derive(Clone, Debug, Encode, Decode, Serialize)]
 pub struct Metadata {
 	pub metadata: RuntimeMetadataLastVersion,
 	pub pallets: BTreeMap<String, PalletMetadata>,
@@ -161,7 +161,7 @@ impl Metadata {
 	}
 
 	#[cfg(feature = "std")]
-	pub fn pretty_format(metadata: &RuntimeMetadataPrefixed) -> Option<String> {
+	pub fn pretty_format<Metadata: Serialize>(metadata: &Metadata) -> Option<String> {
 		let buf = Vec::new();
 		let formatter = serde_json::ser::PrettyFormatter::with_indent(b" ");
 		let mut ser = serde_json::Serializer::with_formatter(buf, formatter);
@@ -171,7 +171,7 @@ impl Metadata {
 }
 
 /// Metadata for a specific pallet.
-#[derive(Clone, Debug, Encode, Decode)]
+#[derive(Clone, Debug, Encode, Decode, Serialize)]
 pub struct PalletMetadata {
 	pub index: u8,
 	pub name: String,
@@ -233,7 +233,7 @@ impl PalletMetadata {
 }
 
 /// Metadata for specific field.
-#[derive(Clone, Debug, Encode, Decode)]
+#[derive(Clone, Debug, Encode, Decode, Serialize)]
 pub struct EventFieldMetadata {
 	name: Option<String>,
 	type_name: Option<String>,
@@ -263,7 +263,7 @@ impl EventFieldMetadata {
 }
 
 /// Metadata for specific events.
-#[derive(Clone, Debug, Encode, Decode)]
+#[derive(Clone, Debug, Encode, Decode, Serialize)]
 pub struct EventMetadata {
 	pallet: String,
 	event: String,
@@ -293,7 +293,7 @@ impl EventMetadata {
 	}
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Encode, Decode)]
+#[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd, Encode, Decode, Serialize)]
 pub struct ErrorMetadata {
 	pallet: String,
 	error: String,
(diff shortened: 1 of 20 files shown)


node-api/src/metadata.rs
```diff
@@ -17,10 +17,8 @@ use frame_metadata::{
 	StorageEntryMetadata, META_RESERVED,
 };
 use scale_info::{form::PortableForm, PortableRegistry, Type};
-use sp_core::storage::StorageKey;
-
-#[cfg(feature = "std")]
 use serde::Serialize;
```
haerdib, line 20: Serde is no_std compatible, so no need to hide behind a std feature only flag.. or was there another reason @echevrier ?

haerdib: Sorry my bad. Seems like serde_derive is necessary, simply using `Serialize` is not enough. You can ignore that one.

src/api/api_client.rs
```diff
@@ -179,29 +191,31 @@ where
...
 
 		match version {
 			Some(v) => serde_json::from_str(&v).map_err(|e| e.into()),
 			None => Err(ApiClientError::RuntimeVersion),
 		}
 	}
 
-	fn _get_metadata(client: &Client) -> ApiResult<RuntimeMetadataPrefixed> {
+	pub fn get_metadata_from_node(client: &Client) -> ApiResult<RuntimeMetadataPrefixed> {
 		let jsonreq = json_req::state_get_metadata();
```
echevrier, line 217: The names `get_metadata`, `get_metadata_from_node` (and similar ones) are a bit confusing, as if the metadata returned by `get_metadata` is not the node's metadata. The difference is not the source of the metadata, but its currentness. 
Wouldn't it be better to have `get_cached_metadata` for `get_metadata` and let `get_metadata` for `get_metadata_from_node` ?

echevrier (commented):
Only one question.

src/api/api_client.rs
haerdib: I see your point, but I think the `get_metadata` should stay as is, because that should be the usual way to retrieve the metadata. Getting the metadata from the node should be the non-normal way, therefore warranting the longer name, I believe.

How about `query_metadata_from_node`?

echevrier: Perfect. That's clear. Thanks

clangenb (commented):
May I ask what was the rationale behind adding metadata as a field?

Spontaneously, I think I would prefer that the Api is stateless to prevent errors of inconsistency.

haerdib:
If the metadata is not chached, that means for every extrinsic sent, the metadata, runtime version and such need to be queried first. Not sure if that's smart?

haerdib:
Just checked: The metadata field has been there since 2019, so I guess I can't answer the original rationale, but I suppose the speed is a concern.

src/api/api_client.rs
```diff
@@ -168,6 +168,18 @@ where
...
+		let metadata = Self::get_metadata_from_node(&self.client).map(Metadata::try_from)??;
+		debug!("Metadata: {:?}", metadata);
+
+		let runtime_version = Self::get_runtime_version_from_node(&self.client)?;
+		info!("Runtime Version: {:?}", runtime_version);
+
+		self.metadata = metadata;
+		self.runtime_version = runtime_version;
+		Ok(self)
+	}
```
clangenb, line 181: I think this should be `pub fn update_runtime(&mut self) -> ApiResult<Metadata>`. The `mut self` is rather used for the builder pattern and not the object itself (btw. I think the `setSigner` method should also be changed).

I don't now if we want to return anything at all, it could also just be `()`

src/api/api_client.rs
clangenb: I think we are mixing concepts here, `get_*` sometimes performs a network request now, and sometimes it does not, this inconsistency is confusing.

I suggest `get_*` always talks to the node, then we don't need to distinguish between `query_metadata_from_node` to emphasize that it actually does the same as most of the `get_*` calls.

We don't have to re-invent the wheel here, we can just do:

`api.metadata()` is a simple getter to the field
`api.get_metdata()` is the network request, however, do we even want to keep it?

Same for the runtime version.

clangenb (commented):
I also have a remark about the naming, but otherwise it looks good. I totally agree about the metadata caching now. :)

src/api/api_client.rs
haerdib: > api.get_metdata() is the network request, however, do we even want to keep it?

Not sure. We need a private node query method, because the api needs it to be able to update the cached metdata, but it does not need to be public.. because one could just run `update_...` and then query the cached metadata.

I don't have a strong opinion here, but well.. why not make it private and force the user to update cached metadata if up-to-date metadata is needed.

src/api/api_client.rs
haerdib: Agreed. Tried to be consistent, with what's there already, but passing a mut reference makes more sense.

src/api/api_client.rs
clangenb: > I don't have a strong opinion here, but well.. why not make it private and force the user to update cached metadata if up-to-date metadata is needed.

This seems good to me! However, the private one should still be `get_metadata` IMO. :)

haerdib: Renamed accordingly 👍

src/api/api_client.rs
haerdib: Done :)

src/api/api_client.rs
haerdib: Also restructured a little, because it was veery chaotic. Updated the PR description accordingly to what has been changed.

echevrier (approved):
Nice work. Thanks

src/api/api_client.rs
```diff
@@ -111,28 +111,93 @@ where
...
 		let multi_signer = MultiSigner::from(pair.public());
 		Some(multi_signer.into_account())
 	}
 
+	/// Get the private key pair of the api signer.
+	pub fn signer(&self) -> Option<&P> {
+		match &self.signer {
+			Some(signer) => Some(signer),
+			None => None,
+		}
```
clangenb, line 147: This can certainly be simplified, no?

clangenb (approved):
I have a question, but it looks very clean to me otherwise!

src/api/api_client.rs
haerdib: Ah well, yeah. Forgot about the as_ref()

PR #339, opened 2022-11-24
Remove substrate versioning and cargo update

### LINKED ISSUE #338: failed to select a version for the requirement `sp-core = "^6.0.0"`
error: failed to select a version for the requirement `sp-core = "^6.0.0"`
candidate versions found which didn't match: 7.0.0
location searched: Git repository https://github.com/paritytech/substrate.git?branch=master
required by package `substrate-api-client v0.7.0 (https://github.com/scs/substrate-api-client.git#b7f07a4e)`

haerdib:
- Removes substrate versioning from .tomls
- `cargo update -p sp-std`

closes #338

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -228,6 +228,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b645a089122eccb6111b4f81cbc1a49f5900ac4666bb93ac027feaecf15607bf"
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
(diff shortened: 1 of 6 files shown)


clangenb (approved):
Nice looks good! On second thought, however, I am uncertain if it is wiser to keep the versions because now it is no longer a simple search and replace if we need to add the versions again because of a -dev suffix.

I let you decide :)

haerdib:
Yeah, I have the same concerns.. But manually updating the version whenever substrate does, also sucks. I think it's worth a try.

PR #340, opened 2022-11-24
Make api and primitives generic over Runtime

haerdib:
- Types defined by the Runtime (Hash, Index, Balance .. ) are now not hardcoded any more, but depend on the Runtime input.
- Renamed generic parameter `P` (probably for Pair) to `Signer`
- Removed pallet dependencies completely from `no_std` due to default wasm compilation.


echevrier (approved):
LGTM

node-api/src/events.rs
```diff
@@ -180,7 +180,7 @@ impl EventDetails {
 
 		// topics come after the event data in EventRecord. They aren't used for
 		// anything at the moment, so just decode and throw them away.
-		let _topics = Vec::<Hash>::decode(input)?;
+		let _topics = Vec::<H256>::decode(input)?;
```
haerdib, line 183: Hardcoded H256 for now.. as they're not used and are unlikely to be different from H256 at the moment.

primitives/src/lib.rs
```diff
@@ -22,20 +22,3 @@ pub use extrinsics::*;
 
 pub mod extrinsic_params;
 pub mod extrinsics;
-
-/// The block number type used in this runtime.
-pub type BlockNumber = u64;
```
haerdib, line 27: Not needed anymore, because all these params should now be generic inside the api-client.

src/api/api_client.rs
```diff
@@ -102,50 +107,56 @@ use std::convert::{TryFrom, TryInto};
...
+	pub fn set_signer(&mut self, signer: Signer) {
 		self.signer = Some(signer);
 	}
 
 	/// Get the public part of the api signer account.
-	pub fn signer_account(&self) -> Option<AccountId> {
+	pub fn signer_account(&self) -> Option<AccountId32> {
 		let pair = self.signer.as_ref()?;
 		let multi_signer = MultiSigner::from(pair.public());
 		Some(multi_signer.into_account())
```
haerdib, line 150: MultiSigner only implements `into_account` for AccountId32. Not sure why, but I guess there's a reason.

clangenb (commented):
Nice to see everything getting so generic. However, I am unsure if depending on the runtime directly is a potential blocker for some use cases, especially the no-std world because it compiles to wasm in no-std. So maybe we need to have a layer in between like:


```rust

pub trait RuntimeConfig {
    type Balance;

    // etc.

   // We could even define the tips here
}

// We could add feature gated configs like:
impl RuntimeConfig for KitchenSinkRuntime {
   type Balance = KitchenSinkRuntime::Balance

   // etc
}

I am not sure if this creates problems on other ends though.

```

examples/staking_payout.rs
```diff
@@ -12,7 +14,7 @@ fn main() {
 	env_logger::init();
 	let from = AccountKeyring::Alice.pair();
 	let client = WsRpcClient::new("ws://127.0.0.1:9944");
-	let mut api = Api::<_, _, PlainTipExtrinsicParams>::new(client).unwrap();
+	let mut api = Api::<_, _, PlainTipExtrinsicParams<Runtime>, Runtime>::new(client).unwrap();
```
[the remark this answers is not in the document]
haerdib: True enough. This example is currently not tested by CI. I'll update it.

haerdib:
You're right, I forgot about the wasm part. But the Config trait itself is not from the Runtime, but from the pallets (e.g. https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L204-L357) . So, worst case, a no_std application would need to implement the Config via your proposed buffer. But the rest should work as is.. or did I overlook something?

So we could offer a `pre-implemented` no-std Runtime as an example, but that could be done in a different issue/PR (first the api needs to be made no-std compatible anyway). Can you agree to that?

clangenb:
The pallets-itself also assume wasm in no-std, so IMO opinion, you can't get around the runtime config trait that I am suggesting. So if you want to merge this PR now, I suggest fixing the trait immediately afterwards, so we don't have a breaking interface in subsequent releases.

primitives/src/types.rs
```diff
@@ -0,0 +1,191 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+*/
+
+//! Re-defintion of substrate primitives.
+//! Needed because substrate pallets compile to wasm in no_std.
+
```
clangenb, line 20: Maybe we can note that all these primitives are from the pallet-balances, aren't they?

src/api/api_client.rs
clangenb: Yeah, because moonbeam does in fact implement an account id with 20 bytes, but but still use the multisigner if I am not mistaken.

src/api/api_client.rs
```diff
@@ -187,38 +197,44 @@ where
...
-	MultiSigner: From<P::Public>,
+	Signer: Pair,
+	MultiSigner: From<Signer::Public>,
 	Client: RpcClient,
-	Params: ExtrinsicParams<Index, Hash>,
+	Params: ExtrinsicParams<Runtime::Index, Runtime::Hash>,
+	Runtime: BalancesConfig,
+	Runtime::Hash: FromHexString,
+	Runtime::Index: From<u32> + Decode,
+	Runtime::Balance: TryFrom<NumberOrHex> + FromStr,
```
clangenb, line 209: We could consider adding those traitbounds to our `FrameSystemConfig` trait or `BalancesConfig` trait. Then we can omit them here.

primitives/src/pallet_traits/pallet_staking_config.rs
```diff
@@ -0,0 +1,99 @@
...
+	type EraPayout = T::EraPayout;
+	type NextNewSession = T::NextNewSession;
+	type MaxNominatorRewardedPerValidator = T::MaxNominatorRewardedPerValidator;
+	type OffendingValidatorsThreshold = T::OffendingValidatorsThreshold;
+	type VoterList = T::VoterList;
+	type TargetList = T::TargetList;
+	type MaxUnlockingChunks = T::MaxUnlockingChunks;
+	type OnStakerSlash = T::OnStakerSlash;
+	type WeightInfo = T::WeightInfo;
+}
```
clangenb, line 99: So we don't offer yet an implementation for no-std for all those pallets? Well, to me this is fine.

primitives/src/types.rs
haerdib: They are not. Should I paste a link from where they are taken?

primitives/src/pallet_traits/pallet_staking_config.rs
haerdib: Not yet, at least. The whole api (except for utils) is not yet no_std compatible. So it just doesn't make sense, since the implementation which are now generic over the Runtime are located in the api / extrinsic.

That will be part of a future PR :)

src/api/api_client.rs
haerdib: True. But I hope to be able to remove most of these once some traits are introduced. Can you bear with that for now?

primitives/src/pallet_traits/pallet_staking_config.rs
clangenb: Alright, this makes sense!

PR #345, opened 2022-11-29
Clean up examples

haerdib:
Rebased work of #242. Thanks goes to @ivanceras .

> - remove the `example_` prefix of all the examples
> - hard code the url endpoint and remove the `cli.yml` and the code for getting the url in each of the examples to make it more lean
> - remove deprecated code
> - Remove the executable permission of some files that don't need to be executable: `Cargo.toml`, `README.md`, license. 
>


echevrier (approved):
LGTM

PR #347, opened 2022-11-29
Update frame-metdata and remove ed25519-zebra patch

### LINKED ISSUE #286: Remove `ed25519-zebra` patch when possible
The patch is necessary because the version used by substrate introduces std leaks in `no_std` when the `full_crypto` feature is enabled (which we do).

* Upstream issue: https://github.com/paritytech/substrate/issues/12225

### LINKED ISSUE #304: Exchange frame-metadata fork with the parity one
Once https://github.com/paritytech/frame-metadata/pull/42 has been merged, change back to the original one.

Might need some updates in the metadata branch.

haerdib:
frame-metadata is now sourced from parity github instead of crates.io (which is very outdated)

closes #286 
closes #304

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ dependencies = [
  "bitvec",
  "derive_more",
  "either",
- "frame-metadata 15.0.0 (git+https://github.com/integritee-network/frame-metadata)",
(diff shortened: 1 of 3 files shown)


clangenb (approved):
Lgtm

PR #354, opened 2022-12-05
Enhance rpc interface and simplify ws client

### LINKED ISSUE #175: `send_extrinsic` should return a hash for all `XtStatus` variants.
#169 introduces a new `XtStatus` to spam tx's even faster. However, the json-rpc interface does not return an xt-hash here.

We should calculate the xt-hash ourselves in that case. However, the stringified interface of `send_extrinsic` makes this a bit tedious. It will get very easy if we change that interface.

So to change that we need:
* change `send_extrinsic` signature to: `send_extrinsics<XT: HexEncode>(xt: XT, exit_on: XtStatus, ...)`
* implement `fn hash(&self) -> H256` for `UncheckedExtrinsicV4`
* return `xt.hash()` in send_extrinsic if the `exit_on == XtStatus::SubmitOnly`

haerdib:
Preparatory step for jsonrpsee inclusion.  

💥  Breaking Changes:
- `send_extrinsic(...)` is exchanged with:
    - `submit_extrinsic`: Replaces the `TxStatus::SubmitOnly`. It directly returns and does not require any subscription. The transaction hash is returned.
    - `watch_extrinsic` and `watch_extrinsic_until`: Requires a subscription, if available, the block hash is returned.
- Subscription functions (`watch_extrinsic`, `subscribe_events` , ..) now handle the channel creation themselves. Instead of giving the `Sender` as an input, the `Receiver` is returned in a `WsSubscriptionWrapper`. The subscription value retrieved via the `next()` function (previously with `value.recv()`). This allows more freedom for the RpcClient implementation, as no mspc channel is enforced.

💚   Feature Changes:
- `RuntimeBlock` is now taken from the Runtime itself instead of a generic function return type.
- RpcClients now implement the function `with_default_url()` and allows external `unsubscription` by the client (via the `WsSubscriptionWrapper`).
- The RpcClients now handles the deserialization, not the api_client.
- Introduces `RpcParams` and the accompanying `rpc_params![..]` macro. This allows to introduce params handling according to the websockets expectation. 
- `json_req.rs` has been removed, as it's not necessary anymore: methods should only get called once per api function and the params are given separately. 
- `subscribe` (previously `start_subscriber`) now expects an additional `unsub` method. This is not yet used, but it allows for better sub / unsub handling in case a more elaborate ws-client is implemented (like jsonrpsee)
- `XtStatus` is now only used as input value. Return value is the substrate equivalent `TransactionStatus`, which can be deserialized directly from the rpc response.
- Heavily cleaned up unnecessary function of the `WsRpcClient` and `TungsteniteClient`

And some other minor corrections..

closes #175


src/rpc/ws_client/client.rs
```diff
@@ -15,172 +15,95 @@
...
+	fn request<R: DeserializeOwned>(&self, method: &str, params: RpcParams) -> Result<R> {
+		let json_req = to_json_req(method, params)?;
+		let response = self.direct_rpc_request(json_req, RequestHandler::default())??;
+		let deserialized_value: R = serde_json::from_str(&response)?;
+		Ok(deserialized_value)
 	}
 }
 
-impl WsRpcClient {
-	pub fn get(
```
haerdib, line 89: Removed a lot of boilerplate code. RpcClient should handle the rpc connection only ( subcribe and send requests), not individual functions such as `watch_until`.

src/rpc/ws_client/mod.rs
```diff
@@ -108,271 +105,17 @@ impl HandleMessage for SubscriptionHandler {
 
 		match value["id"].as_str() {
 			Some(_idstr) => {},
-			_ => {
-				// subscriptions
-				debug!("no id field found in response. must be subscription");
-				debug!("method: {:?}", value["method"].as_str());
-				match value["method"].as_str() {
-					Some("state_storage") => {
```
haerdib, line 116: No need to distinguish between subscription methods - we can simply return the result for any subscription. The differentiation is handled one level below in the `SubscriptionWrapper::next()` where a `Notification` is deserialized.

src/rpc/ws_client/mod.rs
```diff
@@ -108,271 +105,17 @@ impl HandleMessage for SubscriptionHandler {
...
+					out.close(CloseCode::Normal)?;
 				}
 			},
 		};
 		Ok(())
 	}
 }
-
-#[derive(Default, Debug, PartialEq, Eq, Clone)]
-pub struct SubmitOnlyHandler;
```
haerdib, line 146: Obsolete - submit only is a simple request now, that returnes and xt_hash

src/rpc/ws_client/mod.rs
```diff
@@ -108,271 +105,17 @@ impl HandleMessage for SubscriptionHandler {
...
-		debug!("got msg {}", retstr);
-		match result_from_json_response(retstr) {
-			Ok(val) => end_process(out, result, Ok(Some(val))),
-			Err(e) => end_process(out, result, Err(e)),
-		}
-	}
-}
-
-#[derive(Debug, PartialEq, Eq, Clone)]
-pub struct SubmitAndWatchHandler {
```
haerdib, line 167: Obsolete. SubmitAndWatch is now handlded as a simple Subscription that returns a `TransactionStatus`.

echevrier (approved):
Only a small comment.

examples/benchmark_bulk_xt.rs
```diff
@@ -54,7 +53,7 @@ fn main() {
 		);
 		// send and watch extrinsic until finalized
 		println!("sending extrinsic with nonce {}", nonce);
-		let _blockh = api.send_extrinsic(xt.hex_encode(), XtStatus::Ready).unwrap();
+		let _tx_hash = api.submit_extrinsic(xt.hex_encode()).unwrap();
```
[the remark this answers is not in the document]
haerdib: To have one example (and test) with `submit_extrinsic`. As the benchmark is about speed, I guessed that submit_extrinsic is enough anyway.
But the comment above should be adapted

src/rpc/ws_client/client.rs
```diff
@@ -15,171 +15,95 @@
...
+}
 
+impl WsRpcClient {
 	fn start_rpc_client_thread<MessageHandler>(
 		&self,
 		jsonreq: String,
 		result_in: ThreadOut<MessageHandler::ThreadMessage>,
 		message_handler: MessageHandler,
-	) -> Result<()>
+	) -> Result<WsSender>
```
haerdib, line 83: returns sender, to we can externally unsubscribe / shutdown the subscription.

src/rpc/ws_client/subscription.rs
```diff
@@ -0,0 +1,64 @@
...
+impl<Notification: DeserializeOwned> HandleSubscription<Notification>
+	for WsSubscriptionWrapper<Notification>
+{
+	fn next(&mut self) -> Option<Result<Notification>> {
+		let notification = match self.receiver.recv() {
+			Ok(notif) => notif,
+			// Sender was disconnected, therefore no further messages are to be expected.
+			Err(_) => return None,
+		};
+		// Decode to the notification on is expecting.
```
clangenb, line 47: Strictly speaking, you deserialize instead of decoding ;)

src/rpc/ws_client/subscription.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// Support async: #278 (careful with no_std compatibility).
+impl<Notification: DeserializeOwned> HandleSubscription<Notification>
+	for WsSubscriptionWrapper<Notification>
+{
+	fn next(&mut self) -> Option<Result<Notification>> {
+		let notification = match self.receiver.recv() {
+			Ok(notif) => notif,
+			// Sender was disconnected, therefore no further messages are to be expected.
+			Err(_) => return None,
```
clangenb, line 45: Is it ok to not log here?

src/rpc/ws_client/mod.rs
```diff
@@ -108,271 +105,17 @@ impl HandleMessage for SubscriptionHandler {
 
 		match value["id"].as_str() {
 			Some(_idstr) => {},
```
clangenb, line 107: It should not happen, but maybe we want to log here, that we expected a subscription.

src/api/api_client.rs
```diff
@@ -601,23 +544,11 @@ where
...
-	}
-
-	#[cfg(not(feature = "ws-client"))]
-	pub fn send_extrinsic(&self, xthex_prefixed: String) -> ApiResult<Option<Runtime::Hash>> {
+	/// Submit an extrsinic to the substrate node, without watching.
+	pub fn submit_extrinsic(&self, xthex_prefixed: String) -> ApiResult<Runtime::Hash> {
 		debug!("sending extrinsic: {:?}", xthex_prefixed);
-		// XtStatus should never be used used but we need to put something
-		self.client.send_extrinsic(xthex_prefixed, XtStatus::Broadcast)
+		let xt_hash = self.client.request("author_submitExtrinsic", rpc_params![xthex_prefixed])?;
```
clangenb, line 550: 🥳

compose-macros/src/rpc.rs
```diff
@@ -0,0 +1,52 @@
...
+//! It is copied & pasted here to avoid std dependencies.
+
+/// Convert the given values to a [`RpcParams`] as expected by a
+/// Rpc Client (http or websocket).
+///
+/// # Panics
+///
+/// Panics if the serialization of parameters fails.
+#[macro_export]
+macro_rules! rpc_params {
```
clangenb, line 38: Nice, cool macro.

src/api/mod.rs
```diff
@@ -44,14 +44,69 @@ pub use subscription::*;
...
-	/// uses `author_submit` without watching.
-	SubmitOnly = 1,
-	Ready = 2,
-	Broadcast = 3,
+	Ready = 1,
+	Broadcast = 2,
 	InBlock = 4,
-	Finalized = 5,
-	Future = 10,
+	Finalized = 7,
```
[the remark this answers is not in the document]
haerdib: Based on the `TransactionStatus` enum from parity. E.g. XtStatus::Ready u8 = TransactionStatus::Ready u8

src/rpc/ws_client/subscription.rs
haerdib: True enough, and that comment is not nice anyway. Will clean up, thanks for pointing it out 👍

src/rpc/ws_client/subscription.rs
haerdib: Not sure what to log here. The user will know what happened because that's the only way a `None` is returned.

src/api/mod.rs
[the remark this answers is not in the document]
haerdib: We could, but the `TransactionStatus` is a lot more complex. So if the user wants to wait until `InBlock` it'd need an Input
`TransactionStatus<Hash, BlockHash>::InBlock(BlockHash)` or something like that.. don't think that's good.
Another possibility would be to remove the `watch_until` function and instead provide functions such as  `watch_until_in_block`

src/rpc/ws_client/mod.rs
haerdib: Included a warn log

src/rpc/ws_client/subscription.rs
clangenb: Ok, I don't exactly understand the details here. If it is true what you say, I can accept that. :)

src/rpc/ws_client/mod.rs
clangenb: Messi bokup!

src/rpc/ws_client/subscription.rs
```diff
@@ -44,7 +44,7 @@ impl<Notification: DeserializeOwned> HandleSubscription<Notification>
 			// Sender was disconnected, therefore no further messages are to be expected.
 			Err(_) => return None,
 		};
-		// Decode to the notification on is expecting.
+		// Deserialze to the notification one is expecting.
```
clangenb, line 47: Still a typo ;) But honestly, I doubt that the comment is adding any value anyhow.

src/api/mod.rs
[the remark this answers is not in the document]
haerdib: Added a comment - is that understandable?

src/rpc/ws_client/subscription.rs
haerdib: Agreed. Removed

examples/transfer_using_seed.rs
```diff
@@ -58,8 +58,10 @@ fn main() {
 	println!("[+] Composed extrinsic: {:?}\n", xt);
 
 	// Send and watch extrinsic until in block.
-	let tx_hash = api.send_extrinsic(xt.hex_encode(), XtStatus::InBlock).unwrap();
-	println!("[+] Transaction got included. Hash: {:?}\n", tx_hash);
+	let block_hash = api
+		.submit_and_submit_and_watch_extrinsic_until(&xt.hex_encode(), XtStatus::InBlock)
```
clangenb, line 62: find the error: 🕵️

haerdib: No idea what's your problem 🙈

clangenb (commented):
One last comment, looks good otherwise!

examples/benchmark_bulk_xt.rs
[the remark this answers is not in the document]
haerdib: No. submit_extrinsic always returns the tx hash. Submit and watch returns an optional block hash.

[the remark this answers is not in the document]
clangenb: ui, this is confusing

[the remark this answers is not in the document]
haerdib: That's the choice substrate made, so we can't change the output of submit_extrinsic 
(
grafik
)

Though, there's some logic: `submit_extrinsic` does not know in which block the tx get included, as it's a fire and forget. Unlike `submit_and_watch`. Do you think we should change the `submit_and_watch_extrinsic_until` output?

haerdib:
@zhizming-zhong if you're interested: I have updated the tungestenite-client quite a bit. But the functionality should be the same, I tested it locally.

PR #351, opened 2022-12-02
Add jsonrpsee as default ws

haerdib:
closes #133:

For one run of the example `transfer_using_seed`:
- Without jsonrpsee: 
```
2022-12-06 13:50:54 ✨ Imported #109 (0x5083…7ff3)
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:55 Accepting new connection 1/100
2022-12-06 13:50:57 🙌 Starting consensus session on top of parent 0x50835f1edb77c9e2d5f90336bd7e25c581b9cf6051a39f90425d9cb1f37f7ff3
2022-12-06 13:50:57 🎁 Prepared block for proposing at 110 (0 ms) [hash: 0xa57c1ca6755f5fb5e5a9d2bab7d6e4cbe44049400144a04c38850bd3b0d3cdd7; parent_hash: 0x5083…7ff3; extrinsics (2): [0x4d9f…c027, 0x6dda…1895]]
```
- With jsonrpsee:
```
2022-12-06 13:47:09 🔖 Pre-sealed block for proposal at 34. Hash now 0x2346ae46a27fde75d557ca187b634f3de23e25aab109e0876343f2f25e68611d, previously 0xb78844a25c88c5befeb5d91cc84ba675d5902dcd552719ab0097846be12b20b7.
2022-12-06 13:47:09 ✨ Imported #34 (0x2346…611d)
2022-12-06 13:47:09 Accepting new connection 1/100
2022-12-06 13:47:12 🙌 Starting consensus session on top of parent 0x2346ae46a27fde75d557ca187b634f3de23e25aab109e0876343f2f25e68611d
```

PR #355, opened 2022-12-06
Support `tungstenite-rs` rpc client

zhizming-zhong:
This PR

-  implement new rpc client with [tungstenite-rs ](https://github.com/snapview/tungstenite-rs)
-  `tungstenite-client` attempt to reconnect to rpc node once disconnected

Run example with tungstenite
```bash
cargo run --no-default-features --features tungstenite-client,std --example get_storage
```

### DIFF
--- a/src/rpc/tungstenite_client/client.rs
+++ b/src/rpc/tungstenite_client/client.rs
@@ -1,0 +1,330 @@
+/*
+   Copyright 2019 Supercomputing Systems AG
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+	   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+*/
+use crate::{
+	json_req,
+	rpc::{Error as RpcClientError, Result, RpcClient as RpcClientTrait},
+	tungstenite_client::{
+		read_until_text_message, GetRequestHandler, SubmitAndWatchHandler, SubmitOnlyHandler,
+		SubscriptionHandler,
+	},
+	FromHexString, HandleMessage, Subscriber, XtStatus,
+};
+use log::{debug, error, info, warn};
+use serde_json::Value;
+use sp_core::H256 as Hash;
+use std::{
+	fmt::Debug, net::TcpStream, sync::mpsc::Sender as ThreadOut, thread, thread::sleep,
+	time::Duration,
+};
+use tungstenite::{
+	client::connect_with_config,
+	protocol::{frame::coding::CloseCode, CloseFrame},
+	stream::MaybeTlsStream,
+	Message, WebSocket,
+};
+use url::Url;
+
+pub(crate) type MySocket = WebSocket<MaybeTlsStream<TcpStream>>;
+
+#[derive(Debug, Clone)]
+pub struct TungsteniteRpcClient {
+	url: String,
+	max_attempts: u8,
+}
+
+impl TungsteniteRpcClient {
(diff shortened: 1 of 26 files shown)


haerdib:
Awesome, thanks very much. I prefer tungstenite of the outdated ws library anyway. 

FYI: There's currently some work ongoing to support the jsonrpsee client, which handles subscription, connection and stuff in a much nicer way, check out:
https://github.com/scs/substrate-api-client/pull/354 and https://github.com/scs/substrate-api-client/pull/351

But we can merge this PR before the other these two PRs, so you don't have to rebase all that stuff.

src/rpc/tungstenite_client/client.rs
```diff
@@ -0,0 +1,330 @@
...
+						}
+						if socket.can_read() {
+							current_attempt = 1;
+							// After sending the subscription request, there will be a response(result)
+							let msg_from_req = read_until_text_message(&mut socket);
+							match msg_from_req {
+								Ok(msg_from_req) => {
+									debug!("response message: {:?}", msg_from_req);
+									loop {
+										let msg = read_until_text_message(&mut socket);
```
haerdib, line 279: That's a loop within a loop. I'd rather avoid this.

How about getting rid of the loops with one `while` looking something like this:
```rust
while Ok(message) = socket.read_message() {
     match message {
         Message::Text(s) => { // do something .. } ,
         Message::Close(close) => { break; },  // Don't think we should continue the loop if a close message is received
         _ => { debug!("ignore non text message"); },
     }
} 
```
That will make it much more obvious on what's happening within the loop.

src/rpc/tungstenite_client/client.rs
```diff
@@ -0,0 +1,330 @@
...
+										}
+										match message_handler.handle_message(&mut socket) {
+											Ok(msg) =>
+												if let Err(e) = result_in.send(msg.into()) {
+													//thread_message.send(result)
+													error!("failed to send channel: {:?} ", e);
+													return
+												},
+											Err((e, retry)) => {
+												error!("on_subscription_msg: {:?}", e);
```
haerdib, line 292:
```suggestion

```

src/rpc/tungstenite_client/client.rs
```diff
@@ -0,0 +1,330 @@
...
+		let mut socket: MySocket;
+
+		while current_attempt <= self.max_attempts {
+			match connect_with_config(url.clone(), None, 2) {
+				Ok(res) => {
+					socket = res.0;
+					let response = res.1;
+					debug!("Connected to the server. Response HTTP code: {}", response.status());
+					if socket.can_read() {
+						current_attempt = 1;
```
haerdib, line 77: That's the second allocation to 1

src/rpc/tungstenite_client/client.rs
```diff
@@ -0,0 +1,330 @@
...
+		let mut current_attempt: u8 = 1;
+		let mut socket: MySocket;
+
+		while current_attempt <= self.max_attempts {
+			match connect_with_config(url.clone(), None, 2) {
+				Ok(res) => {
+					socket = res.0;
+					let response = res.1;
+					debug!("Connected to the server. Response HTTP code: {}", response.status());
+					if socket.can_read() {
```
haerdib, line 76: This code is getting hard to read because a lot of ifs and loops are used. To avoid stacking `if { .. if { .. ` you could do the opposite, look for an error and return the error if found.
E.g. simply do 
```rust
socket.can_read()?;
// Check for ping message
socket.read_message()?; 
```
If you want to handle the error in a specific way, you could also write these checks in a separate function and handle the error here. Like
```rust
if let Err(e) = check_connection(&mut socket) {
    error!("adbds");
}
// continue..
```

src/rpc/tungstenite_client/client.rs
haerdib: I also think that's a lot of dubplicate code. Maybe you could extract a function and reuse it in the message handlers.

zhizming-zhong:
I had updated the code, please check.

haerdib (approved):
Thanks for working on it! It looks better now. As it does not change anything in the current ws-client I think that's good to go :)

There will be changes incoming due to the RPC interface adaption in #354 , but that should not break the core of this client.

PR #356, opened 2022-12-06
Add api traits

### LINKED ISSUE #267: Abstract api-client main features behind a trait

haerdib:
closes #267


haerdib:
closed in favour of #363

PR #359, opened 2022-12-07
updated and added usage of ajuna network

darkfriend77:


clangenb (approved):
Nice, thanks for the addition!

Issue #357, opened 2022-12-06
Don't require `Pair` and `ExtrinsicParams` trait bound for getter implementations.


clangenb:
We always require `Pair` or `ExtrinsicsParams` for all the `Api` implementations. This leads to the weird situation, where an actual pair needs to be supplied even if we never set a signer, as we don't need to sign anything, e.g.:

https://github.com/scs/substrate-api-client/blob/7ff294c406f0c03cd13b08e59272f6a3c0675836/examples/get_blocks.rs#L31

If we remove the trait bound for all getters, we can instantiate the `Api` like this, which is much simpler:

```rust
let api =  Api::<(), _, ()>::new(client).unwrap()
```

haerdib:
Jup jup,  you're right, my bad !

haerdib:
The Pair problem should be solved with PR #356. The ExtrinsicParams will need to be solved in a separate PR

haerdib:
If we separate, then I believe two are enough. The ExtrinsicSender needs to be able to `get` things as well. At least to know its nonce. Therefore, I'd vote for `GetterClient` and `FullApiClient`. Three would just complicate the matter.

Hm.. maybe yes to the Config struct. But maybe both make sense, as the GetterClient would be a lot simpler than the FullApiClient, not needing any node information at all. 

If it's alright with you, I'll create a follow-up issue with a `GetterClient` creation and close this one. The Config issue already exists.

clangenb:
This is ok for me, however, my ideas were just a brainstorm, take it with a grain of salt.

haerdib:
Don't worry. I was thinking about that as well. And I think subxt has a similar concept with offline and online client.

PR #363, opened 2022-12-08
Add api traits

### LINKED ISSUE #267: Abstract api-client main features behind a trait

haerdib:
Adds traits to the api and removes obsolete impl enforcements on the Api where possible (see #357)

closes #267


src/api/error.rs
```diff
@@ -33,15 +33,14 @@ pub enum Error {
 	MetadataFetch,
 	#[error("Operation needs a signer to be set in the api")]
 	NoSigner,
-	#[error("RpcClient error: {0:?}")]
+	#[error("RpcClient  error: {0:?}")]
```
clangenb, line 36: typo

src/api/interfaces/frame_system.rs
```diff
@@ -0,0 +1,171 @@
+/*
```
clangenb, line 1: Minor naming convention: in rust you talk about traits, not interfaces.

Second, with this name, I would expect only the trait declarations, not the implementations here.

src/api/interfaces/generic_storage.rs
```diff
@@ -0,0 +1,255 @@
+/*
```
clangenb, line 1: I would just call it `storage` the name already implies generic.

src/api/interfaces/frame_system.rs
clangenb: Maybe call it `rpc_implementations`?

src/api/interfaces/pallet_balances.rs
```diff
@@ -0,0 +1,106 @@
+/*
```
clangenb, line 1: Hmm, `pallet_balances` is wrong actually, only the existential deposit belongs to the pallet balances. The rest is `pallet-transaction-payment`

src/api/interfaces/frame_system.rs
```diff
@@ -0,0 +1,171 @@
...
+	Runtime: FrameSystemConfig,
+	Params: ExtrinsicParams<Runtime::Index, Runtime::Hash>,
+	Runtime::Header: DeserializeOwned,
+	Runtime::Hash: FromHexString,
+{
+	type Header = Runtime::Header;
+
+	fn get_finalized_head(&self) -> ApiResult<Option<Runtime::Hash>> {
+		let finalized_block_hash =
+			self.client().request("chain_getFinalizedHead", rpc_params![])?;
```
clangenb, line 115: this is obviously not from the `frame-system-pallet`

clangenb (commented):
Finally, the monolith has been separated!

However, I think that the separation is not really coherent with the substrate structure. What I propose to cleanly separate the stuff:

```
rpc_implementatsions/
   chain
   state
   constants
   frame_system
   pallet-balances
   pallet-transaction-payment
```
I do wonder if we should split up the subscriptions and put them into the respective domain, e.g.,

`subscribe_finalized_head` -> `chain` is this queries the `chain` RPC of the substrate node.

On the other hand, this would introduce a bit more complexity. I am unsure.

src/api/interfaces/frame_system.rs
haerdib: I was not thinking about rust interfaces (e.g. traits) but rather the node interface. But, agreed, it is confusing.

haerdib: How about `rpc_interfaces`? It's not really an implementation, I think. Only a helper for easy rpc access and forwarding it to the client.

haerdib: Or `rpc_helpers`?

src/api/interfaces/generic_storage.rs
haerdib: Renamed to `state`

src/api/interfaces/pallet_balances.rs
haerdib: You're right, moved :)

src/api/interfaces/frame_system.rs
haerdib: Moved to chain.rs

src/api/interfaces/subscription.rs
```diff
@@ -0,0 +1,104 @@
...
+};
+pub use ac_node_api::{events::EventDetails, StaticEvent};
+use ac_node_api::{DispatchError, Events};
+use ac_primitives::{ExtrinsicParams, FrameSystemConfig};
+use log::*;
+use serde::de::DeserializeOwned;
+use sp_core::storage::StorageChangeSet;
+
+// FIXME: This should rather be implemented directly on the
+// Subscription return value, rather than the api. Or directly
```
haerdib, line 26: I don't like how it's currently done. But changing this would break the interface. So that should be done in a different PR

haerdib (commented):
Moved all subscription, where possible, to the correspondent file. I think it makes more sense than to have all subscriptions in one file.

src/api/interfaces/generic_storage.rs
clangenb: Thanks!

examples/event_callback.rs
```diff
@@ -38,7 +38,7 @@ async fn main() {
 		Api::<sr25519::Pair, _, AssetTipExtrinsicParams<Runtime>, Runtime>::new(client).unwrap();
 
 	println!("Subscribe to events");
-	let mut subscription = api.subscribe_events().unwrap();
+	let mut subscription = api.subscribe_system_events().unwrap();
```
clangenb, line 41: Good renaming!

src/api/interfaces/pallet_balances.rs
clangenb: Yaay! 🥳

src/api/interfaces/state.rs
```diff
@@ -0,0 +1,291 @@
...
+use ac_compose_macros::rpc_params;
+pub use ac_node_api::{events::EventDetails, StaticEvent};
+use ac_primitives::{ExtrinsicParams, FrameSystemConfig};
+use codec::{Decode, Encode};
+use log::*;
+use serde::de::DeserializeOwned;
+use sp_core::storage::{StorageChangeSet, StorageData, StorageKey};
+
+/// Generic interface to substrate storage.
+pub trait GetState<Hash> {
```
clangenb, line 23: I would change this to `GetStorage`

src/api/interfaces/frame_system.rs
clangenb: I would still vote for `rpc_implementations` they are not helpers for the RPC, they are actual (client) implementations in my opinion. Or maybe `rpc-api`?

src/api/interfaces/subscription.rs
clangenb: Now, we only have events here. Maybe we can call it `EventSubscription`?

src/api/interfaces/subscription.rs
```diff
@@ -0,0 +1,104 @@
...
+// subcribe. Should be looked at in #288
+// https://github.com/scs/substrate-api-client/issues/288#issuecomment-1346221653
+pub trait EventSubscriptionHelper<Client, Hash>
```
clangenb, line 29: I would drop the helper here.

src/api/interfaces/state.rs
haerdib: There's already a GetStorage in node_api to retrieve metadata storage values. I'd rather avoid having two equally named storage.

Two options, if you don't like `GetState`:
- rename the `GetStorage` from node api to `GetStorageTypes`
- rename `GetState` to `GetStateStorage` 

or similar naming.. what do you think?

src/api/interfaces/subscription.rs
haerdib: Ah, actually, following the rust convention, that should rather be a verb noun. How about `SubscribeEvents`?

src/api/interfaces/frame_system.rs
haerdib: 👍 rpc-api

src/api/interfaces/state.rs
haerdib: Went for the first one now - I think `GetStorageTypes` makes more sense anyhow.

clangenb: Yes, I agree! 👍

src/api/interfaces/subscription.rs
clangenb: Yes, I agree that is even better, but the last thing would to also rename the file accordingly to `subscribe_events` :)

haerdib: Done :)

clangenb (approved):
Whoopwhoop!

PR #368, opened 2022-12-09
Add new test crate

### LINKED ISSUE #353: Introduce a test suite for system testing
Currently, the provided examples are used for system testing. But to have a full-suit system test, more tests (but less verbose) should be introduced. That way, more functionality could be tested, without replicating everything.

Current examples should still be tested in CI, to make sure they are up to date. 

Open problem: How to provide (CI tested) examples that use the PlainTip? Nightly job with substrate node template (which is not provided as docker image currently)?

haerdib:
Test the full functionality of the api client.

Update CI build: 
- replace `--all` (deprecated) with `--workspace` (see https://doc.rust-lang.org/cargo/commands/cargo-build.html) 
- remove example build. Examples are built in `cargo test` anyway.
- Add some better namings to jobs / steps.

Not CI related:
- Remove example `get_existential_deposit`. That's not really an useful example and is covered by the tests now.

closes  #353


testing/bin/subscription_tests.rs
```diff
@@ -0,0 +1,91 @@
...
+struct ExtrinsicSuccess {
+	_dispatch_info: DispatchInfo,
+}
+
+impl StaticEvent for ExtrinsicSuccess {
+	const PALLET: &'static str = "System";
+	const EVENT: &'static str = "ExtrinsicSuccess";
+}
+
+#[tokio::main]
```
clangenb, line 38: I did not notice this before, we need the async runtime because jsonrpsee uses it internally?

haerdib: Yes, exactly.

testing/examples/author_tests.rs
```diff
@@ -0,0 +1,98 @@
+/*
```
clangenb, line 1: Are these now tests or examples :D? I suggest that we stay consistent with the naming. We can go either for tests or examples, but I am against this schizophrenic scheme.

clangenb: Actually i think we should either make it bins, then we don't need to add the `--example` flag all the time, what do you think?

clangenb (commented):
Nice cleanup thanks! However, I have some questions. Also, I believe that this PR solves different issues at the same time. Is there a reason that you can't separate this PR?

primitives/src/types.rs
```diff
@@ -91,6 +91,7 @@ impl<Balance: AtLeast32BitUnsigned + Copy> InclusionFee<Balance> {
 pub struct FeeDetails<Balance> {
 	/// The minimum fee for a transaction to be included in a block.
 	pub inclusion_fee: Option<InclusionFee<Balance>>,
+	#[cfg_attr(feature = "std", serde(skip))]
```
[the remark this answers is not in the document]
haerdib: Copy pasted from substrate.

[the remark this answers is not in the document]
clangenb: So weird, can you add maybe that this mirrors the substrate implementation?

testing/examples/author_tests.rs
haerdib: The good thing about examples is that they are built automatically with `cargo test` (but not run). 

Binaries are also possible. Implication: Binaries are built with normal "cargo build --release". Not sure if that's a drawback or not.
What's certainly a drawback: CI needs to be adapted quite a bit, because the binaries are stored in the `release` folder. Or is there a way to define an output folder?

primitives/src/types.rs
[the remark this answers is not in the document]
haerdib: That's already described in the comment above (including the link). Or what would you like?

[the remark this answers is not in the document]
clangenb: haha, sorry. I almost had the feeling then I wrote that. Sorry for the inconvenience.

testing/examples/author_tests.rs
clangenb: Nope, the change would imply that we need to copy it from another folder. Hmm, it is hard to say.

Ok, I will back off. If you think this is the suitable solution, I happily go with it.

primitives/src/types.rs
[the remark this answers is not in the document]
haerdib: Haha, no inconvenience at all. Thanks for looking over my code and even more so for making sure that it's documented 😻

testing/examples/author_tests.rs
haerdib: Suitable... not sure. But the simplest one at leat 😄

clangenb (approved):

echevrier (approved):
LGTM

PR #369, opened 2022-12-13
Add example to separate crate

haerdib:
Reasoning: cleaner separation between examples and actual crate. Removes unnecessary feature inside the examples and shows how to import the api-client inside the examples/Cargo.toml

- Clean up of .tomls from unused dependencies


haerdib:
working PR: #371

PR #371, opened 2022-12-14
Move exampels to separate crate

haerdib:
This way, the examples don't need to inherit all the features of the api-client.  They import the api-client with the necessary features. This serves as a better example on how to import and use the api-client.


echevrier (commented):
I don't like the repetition examples/examples. Instead we could regroup the examples:
examples/
- compose_extrinsic/
- events/
- storage/
...

haerdib:
That sadly doesn't work because of cargo. To name the second example folder `examples` is necessary to mark them as `examples`. Otherwise, each examples of them would need to be defined in the .toml file as `[[example]]` (see https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples)

echevrier (approved):
Then :)

examples/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+frame-support = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+frame-system = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+kitchensink-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+pallet-balances = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+pallet-identity = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+pallet-staking = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-core = { features = ["full_crypto"], git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-keyring = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-runtime = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-version = { git = "https://github.com/paritytech/substrate.git", branch = "master" }
```
clangenb, line 25: I like that these deps no longer live in the root workspace.

clangenb (approved):
Nice, lgtm!

PR #372, opened 2022-12-14
Clean up tomls

haerdib:
- removes unused deps
- sorts deps according to no_std and std / wasm only
- update env_logger to v0.10.0


clangenb (commented):
LGTM

echevrier (approved):
Nice

PR #374, opened 2022-12-15
Fix bugs and remove unused Clone / Debug enforcements

haerdib:
Fixes several bugs noticed in #368:
- adds missing serde skips in primitives types.  These values are not serialized on substrate side, so the deserialization on the api client side needs to skip them as well.
- `XtStatus::Finalized` now matches the number of `TransactionStatus`, which is 6. Otherwise, one would have waited forever until Finalized Status.
- Removes unnecessary enforcement of Decode and Clone on unused `V` of `get_storage_map_proof` and `get_storage_double_map_proof`


clangenb (approved):
Beautiful, thanks for the extraction of these changes!

clangenb (approved):

echevrier (approved):
LGTM

PR #375, opened 2022-12-15
Make api no_std compatible and clean up pub re-exports

haerdib:
❗ Breaking changes: some imports need to be imported differently now. Previously, everything has been pub in a quite unordered manner. This is now a little more ordered:
- everything in the `api` folder can be imported like usual directly with `substrate-api-client::xxx`
- the rpc folder now needs to be specified accordingly. E.g. `substrate-api-client::rpc::xxx`
With this exports of same named struct / traits is avoided (like `Error`).
- Most public reexports have been deleted. Reasoning: Because clippy does not detect unused pubs, they were growing very outdated and were not even used any more in the api-client (like the `json::Value`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4793,16 +4793,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "sp-rpc"
-version = "6.0.0"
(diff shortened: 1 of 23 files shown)


haerdib:
Cosing in favour of #377 and #376

PR #376, opened 2022-12-15
Clean up pub exports

haerdib:


src/api/error.rs
```diff
@@ -21,7 +21,7 @@ use ac_node_api::{
 	DispatchError,
 };
 
-pub type ApiResult<T> = Result<T, Error>;
+pub type Result<T> = core::result::Result<T, Error>;
```
clangenb, line 24: I agree so much about this change!

clangenb (approved):
Nice, this looks good! Much needed cleanup!

Issue #279, opened 2022-09-12
Be fully `no_std` compatible (except for ws rpc)


haerdib:
Would include hosting a ws client in no-std. To my knowledge, this is not yet available as a library.

clangenb:
We do not necessarily need a ws-client, just any networking interface that implements the [RpcClient](https://github.com/scs/substrate-api-client/blob/b91d78b53fdc3a05b3c1d1e7fe0fd6b8cfc0dd17/src/std/rpc/ws_client/client.rs#L37) or the [subscriber](https://github.com/scs/substrate-api-client/blob/b91d78b53fdc3a05b3c1d1e7fe0fd6b8cfc0dd17/src/std/rpc/ws_client/mod.rs#L60).

The cool thing is, some `no_std` libraries provide their own networking abstractions, like we have it in SGX. Hence, it is already a very useful thing to provide everything in `no_std` except the ws-client.

As the main part is already generic over the the traits above, we should be able to make almost everything `no_std` compatible by now.

haerdib:
I have looked into it a little. Making the api-client no_std compatible is not that much work at all, if one leaves out the ws-client. If #267 is done, I think most of it should be solved already. If one leaves out the ws-client.

Regarding the ws-client:
I haven't found anything (once again) that is no_std even nearly `no_std` compatible. [tungstenite](https://github.com/snapview/tungstenite-rs) seems to be a fairly simple one, so one could look into that to see how much needs to done to make it `no_std` compatible. Maybe even a from scratch implementation is necessary. But what's certain: a `no_std` client will need to be fairly simple and slow. No async, no multithreading, no mio. And a lot of work. Not sure if that's worth it. Maybe a fire-and-forget client is somewhat more reasonable.

clangenb:
Networking does usually need the very low-level primitives that interface with the hardware. I don't actually expect a generic `no_std` websocket client to be available at all. It is much more likely that the `ws_client` will be supplied by the user of the library, like our sgx compatible one, or one that is supplied by a browser lib.

Issue #317, opened 2022-11-14
Clean up errors: use `this_error` in `no_std` or remove completely


haerdib:
Tracking issue: https://github.com/dtolnay/thiserror/issues/196

haerdib:
A new crate has been published: https://crates.io/crates/thiserror-core

haerdib:
Investigate if it should actually be used: 

> Just remember, if IoT should be one of the target deployments of this, we might want to get rid of `thisError` in no_std whatsoever, at least for production builds. The footprint of the error-chain that errors contain is not negligible.

https://github.com/scs/substrate-api-client/pull/384#pullrequestreview-1224028905

And if not, what else should be used, and maybe clean up completely.

PR #384, opened 2022-12-19
Make api/rpc interfaces no_std compatible

haerdib:
One step further towards `no_std` compatibility:
- `compsoe-macros` is now fully `no_std` compatible
- `api/rpc` is now fully `no_std` compatible
- remove unused `primitives-types` import in `Cargo.toml`

Drawbacks:
- `thiserror` needed to be changed to `thiserror-core`. But that should be temporary only. And better than removing it completely (I hope).
- Some errors needed to be changed from Error-Types to Strings, because they are not available in `no_std`.

one step towards  https://github.com/scs/substrate-api-client/issues/279

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -15,11 +15,7 @@
 
 */
 
-#[cfg(not(feature = "std"))]
-extern crate alloc;
-#[cfg(not(feature = "std"))]
 use alloc::{string::String, vec::Vec};
-
 use hex::FromHexError;
 use sp_core::{storage::StorageKey, twox_128, H256};
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5188,7 +5188,6 @@ dependencies = [
  "kitchensink-runtime",
  "log",
  "parity-scale-codec",
- "primitive-types",
  "serde",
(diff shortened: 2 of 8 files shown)


src/utils.rs
```diff
@@ -15,11 +15,7 @@
 
 */
 
-#[cfg(not(feature = "std"))]
```
haerdib, line 18: No need to use alloc in `no_std` only. `std::string::String` points to `alloc::string::String` as well. So I removed this completely.

echevrier (approved):
LGTM

clangenb (approved):
Nice looks good!

Just remember, if IoT should be one of the target deployments of this, we might want to get rid of `thisError` in no_std whatsoever, at least for production builds. The footprint of the error-chain that errors contain is not negligible.

haerdib:
Good point. But maybe a full cleanup would be sensible, as propagating Strings is also not very efficient, I suppose. I commented in https://github.com/scs/substrate-api-client/issues/317 accordingly.

clangenb (approved):
LGTM!

Issue #288, opened 2022-10-07
`send_extrinsic(xt,  XtStatus::Finalized)` does not report error upon xt failure


haerdib:
`.send_extrinsic(xt.hex_encode(), XtStatus::Finalized)`

Gives back the hash even though the extrinsic has failed. Is there a way to prevent that?

haerdib:
This is not a bug actually, because the extrinsic, even when failed, gets finalized. However, just like in subxt, we could provide a functionality  `check_success` or similar. This function would then check the associated events of this call and check if there's been an `ExtrinsicFailed` event or `ExtrinsicSuccess`.

Example from subxt: 

```rust
pub async fn wait_for_success(
        &self,
    ) -> Result<crate::blocks::ExtrinsicEvents<T>, Error> {
        let events = self.fetch_events().await?;

        // Try to find any errors; return the first one we encounter.
        for ev in events.iter() {
            let ev = ev?;
            if ev.pallet_name() == "System" && ev.variant_name() == "ExtrinsicFailed" {
                let dispatch_error =
                    DispatchError::decode_from(ev.field_bytes(), &self.client.metadata());
                return Err(dispatch_error.into())
            }
        }

        Ok(events)
    }
```

haerdib:
Also take a look at the function `wati_for_event`. The current implementation is not very statisfactory because the parameter subscription needs to be passed around too much.

Issue #90, opened 2020-05-21
use Bytes type in calls that send transactions


clangenb:
We always hex encode extrinsics on our own before sending them.

we could use `sp_core::Bytes` type as parameter type  to the json requests = `Struct Bytes(Vec<u8>)`, which has custom serialization that prepends `0x`. Then we can simply pass the encoded value of the extrinsic

clangenb:
Actually, I think we should just add a method (don't remove the current one accepting a string):

`fn send_xt<C: Encode>(xt: UncheckedExtrinsic<C>, xt_status: XtStatus)`

This will remove some boilerplate code on the caller site.

haerdib:
I think the validation should happen during the call creation, not the sending. See https://github.com/scs/substrate-api-client/issues/416

What do you think?

PR #386, opened 2022-12-20
SubmitAndWatch: Return `ExtrinsicReport` instead of block hash

haerdib:
❗ Breaks `submit_and_watch_extrinsic` and `submit_extrinsic` interfaces: 
- extrinsic submission now expects an  encoded extrinsic. Hex encoding is no longer allowed / required. The hex encoding is done in the functions themselves. This is needed to calculate the extrinsic hash.
- The return value of `submit_and_watch_extrinsic` has been updated to `ExtrinsicReport` which includes the extrinsic hash, the (maybe) block hash,  the transaction status and, possibly, the associated events. This has the following advantages:
    - Clearly distinguishes between the xt hash (returned by `submit_extrinsic`) and the block hash previously returned.
    - Allows to return all associated items of the extrinsic to the caller.

Preparatory step for #288.


src/api/mod.rs
```diff
@@ -23,8 +23,37 @@ pub mod api_client;
...
+	pub extrinsic_hash: Hash,
+	// Block hash of the block the extrinsic was included in.
+	// Only available if watched until at least `InBlock`.
+	pub block_hash: Option<Hash>,
+	// Last known Transaction Status.
+	pub status: TransactionStatus<Hash, Hash>,
+	// Events assosciated to the extrinsic.
+	// Only available if explicitly stated, because
+	// extra node queries are necessary to fetch the events.
+	pub events: Option<Vec<EventDetails>>,
```
haerdib, line 43: That's not yet used, but will be soon with the incoming `watch_until_success` function.

echevrier (approved):
LGTM

src/api/rpc_api/chain.rs
```diff
@@ -20,8 +20,7 @@ use ac_compose_macros::rpc_params;
 use ac_primitives::{ExtrinsicParams, FrameSystemConfig};
 use log::*;
 use serde::de::DeserializeOwned;
-use sp_core::Pair;
-use sp_runtime::{generic::SignedBlock, traits::GetRuntimeBlockType, MultiSignature};
+use sp_runtime::{generic::SignedBlock, traits::GetRuntimeBlockType};
```
clangenb, line 23: 🥳

src/api/rpc_api/mod.rs
```diff
@@ -23,3 +24,7 @@ pub mod pallet_balances;
 pub mod pallet_transaction_payment;
 pub mod state;
 pub mod subscribe_events;
+
+pub(crate) fn extrinsic_has_failed(event_details: &EventDetails) -> bool {
+	event_details.pallet_name() == "System" && event_details.variant_name() == "ExtrinsicFailed"
+}
```
clangenb, line 30: Why does this not go into the `ac_node_api`?

src/api/rpc_api/author.rs
```diff
@@ -32,7 +34,7 @@ pub trait SubmitExtrinsic {
 
 	/// Submit an extrsinic to the substrate node, without watching.
 	/// Retruns the extrinsic hash.
-	fn submit_extrinsic(&self, xthex_prefixed: String) -> Result<Self::Hash>;
+	fn submit_extrinsic(&self, encoded_extrinsic: Vec<u8>) -> Result<Self::Hash>;
```
clangenb, line 37: I have also thought about adding a function like this, https://github.com/scs/substrate-api-client/issues/90:

```
fn send_xt<C: Encode>(xt: UncheckedExtrinsic<C>) -> Result<Self::Hash>
```

Have you considered this? But this would be an addition, and not a replacement.

src/api/rpc_api/author.rs
```diff
@@ -60,51 +63,58 @@ where
...
 }
 
 impl<Signer, Client, Params, Runtime> SubmitAndWatch<Client, Runtime::Hash>
 	for Api<Signer, Client, Params, Runtime>
 where
 	Client: Subscribe,
 	Params: ExtrinsicParams<Runtime::Index, Runtime::Hash>,
-	Runtime: FrameSystemConfig,
+	Runtime: FrameSystemConfig + GetRuntimeBlockType,
+	Runtime::RuntimeBlock: BlockTrait + DeserializeOwned,
```
clangenb, line 85: Why is the `RuntimeBlock` traitbound needed?

src/api/mod.rs
clangenb: Do you plan that one can submit explicit events that we should wait for? I ask because, we can't know at this level, which events are associated to an extrinsic. We only know which ones are in this block.

src/api/mod.rs
```diff
@@ -23,8 +23,37 @@ pub mod api_client;
 pub mod error;
 pub mod rpc_api;
 
+use ac_node_api::EventDetails;
 use serde::{Deserialize, Serialize};
 
+/// Extrinsic report returned upon a submit_and_watch request.
+/// Holds as much information as available.
+#[derive(Debug, Clone)]
+pub struct ExtrinsicReport<Hash> {
```
clangenb, line 32: We should definitely add decode and encode here, maybe also serialize and deserialize. :)

clangenb (commented):
Looks very nice and clean, I like it. I have just some questions. :)

src/api/mod.rs
haerdib: Yes, the extrinsics are retrieved from the block, but they can be sorted to `associated events` with the phase:
```rust
// Filter events associated to our extrinsic.
		let associated_event_results = events.iter().filter(|ev| {
			ev.as_ref()
				.map_or(true, |ev| ev.phase() == Phase::ApplyExtrinsic(xt_index as u32))
		});
```
polkadot.js does the same :

grafik

src/api/rpc_api/author.rs
haerdib: I thought about it, but this enforces UncheckedExtrinsic. I thought having it encoded still leaves the same freedom as the String one. But I agree, having one that does enforce a specific Extrinsic type would be good, as it is less error-prone.

src/api/rpc_api/author.rs
haerdib: Ah, that's not yet necessary, my bad. Good catch!

src/api/mod.rs
haerdib: Agreed, I'll add it for now!

src/api/rpc_api/mod.rs
haerdib: good point, let me try that

src/api/rpc_api/author.rs
clangenb: Yes, I like enforcing these things. This is what rust is about for me! :D

src/api/mod.rs
clangenb: Oh wow, I did not know about that feature, this is very cool!

src/api/rpc_api/author.rs
haerdib: Aye, Rust is, but Substrate is less so. Respective our client. In case the Extrinsic Type is updated, our client will be rendered useless until updated. So having both is a good compromise I think.

src/api/mod.rs
clangenb: Ok, fine with me.

src/api/rpc_api/author.rs
clangenb: Substrate enforces a lot too IMO.

Whatever, I agree having both should be the best.

src/api/rpc_api/subscribe_events.rs
```diff
@@ -75,11 +75,7 @@ where
 				// Check for failed xt and return as Dispatch Error in case we find one.
 				// Careful - this reports the first one encountered. This event may belong to another extrinsic
 				// than the one that is being waited for.
-				if extrinsic_has_failed(&event_details) {
-					let dispatch_error =
-						DispatchError::decode_from(event_details.field_bytes(), self.metadata());
-					return Err(Error::Dispatch(dispatch_error))
-				}
+				event_details.check_failed()?;
```
clangenb, line 78: I am uncertain if I agree with the naming. I propose a few here:

* check_if_failed
* xt_successful
* xt_failed

What do you think?

clangenb (commented):
Very nice, I only have a comment left.

src/api/mod.rs
haerdib: https://github.com/scs/substrate-api-client/issues/397

src/api/rpc_api/subscribe_events.rs
haerdib: I was thinking about the first one as well. Let me adapt.

clangenb (approved):
Thanks, looks good to me now!

PR #390, opened 2022-12-21
Udpate to newest substrate master commit

haerdib:


primitives/src/pallet_traits/frame_system_config.rs
```diff
@@ -58,7 +58,6 @@ pub trait FrameSystemConfig {
 		+ Copy
 		+ sp_std::hash::Hash
 		+ sp_std::str::FromStr
-		+ MaybeMallocSizeOf
```
haerdib, line 61: Was removed it seems, check out: https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L252-L253

echevrier (approved):
LGTM

PR #394, opened 2022-12-22
Add helper methode api.compose_extrsinic_offline

### LINKED ISSUE #195: introduce `offline_xt` method
In the encointer client, we have defined this useful helper:

We should also introduce a getter for these fields, so we don't need to clone.

```
/// Wrapper around the `compose_extrinsic_offline!` macro to be less verbose.
pub fn offline_xt<C: Encode + Clone>(
	api: &Api<sr25519::Pair, WsRpcClient>,
	call: C,
	nonce: u32,
) -> UncheckedExtrinsicV4<C> {
	compose_extrinsic_offline!(
		api.clone().signer.unwrap(),
		call,
		nonce,
		Era::Immortal,
		api.genesis_hash,
		api.genesis_hash,
		api.runtime_version.spec_version,
		api.runtime_version.transaction_version
	)
}
```

I think this will be of use here too.

haerdib:
Makes the offline extrinsic creation less verbose.

closes #195


clangenb (approved):
Nice!

clangenb (approved):

PR #403, opened 2022-12-30
Add `get_storage_keys_paged`

### LINKED ISSUE #227: Which is the proper query to call query_storage_map and get an iterable?
If i have a storage_map_query and i want to omit the option i usually get a vac of tuples through JS, i tried passing Vec<(K,V)> to the query storage map method but i always get a storage decode error, which is the proper way of getting a paginated result?

haerdib:
closes #227


clangenb (approved):
I was actually waiting for this!

PR #405, opened 2022-12-30
Bh/400 clean up extrinsic params builder

haerdib:
better?


src/api/api_client.rs
```diff
@@ -298,8 +299,8 @@ mod tests {
 			create_mock_api(genesis_hash, runtime_version.clone(), metadata, Default::default());
 
 		// Information for Era for mortal transactions.
-		let builder = PlainTipExtrinsicParamsBuilder::<Runtime>::new();
-		api.set_extrinsic_params_builder(builder);
+		let builder = PolkadotAdditionalParams::new();
```
Niederb, line 302: Rename the variable?

src/api/api_client.rs
```diff
@@ -154,19 +154,20 @@ where
 	}
 
 	/// Set the extrinscs param builder.
```
Niederb, line 156: Adjust the comment

src/api/api_client.rs
```diff
@@ -154,19 +154,20 @@ where
...
-	pub fn set_extrinsic_params_builder(&mut self, extrinsic_params: Params::OtherParams) {
-		self.extrinsic_params_builder = Some(extrinsic_params);
+	pub fn set_additional_params(&mut self, extrinsic_params: Params::AdditionalParams) {
+		self.additional_extrinsic_params = Some(extrinsic_params);
 	}
 
 	/// Get the extrinsic params, built with the set or if none, the default Params Builder.
```
Niederb, line 161: Adjust the comment

Niederb (commented):
LGTM, just a few nitpicky suggestions. 
Regarding `Polkadot` vs `SubstrateDefault` I don't have a strong opinion. I think I would leave it as is for now.

clangenb:
Nice looks in general good great that you noticed that we can remove the specific builder definitions! I have a few remarks about the terminology:

> Renamed Base or Default to Polkadot because they are polkadot-node compatible (and default substrate node). Base did not fit IMO because that implicates that it is extensible. But it is not. One would need to redefine the Params in case something is changed on the node side.

`Base` might be a bit off, but I think polkadot is wrong because it implies that the params are less generic than they actually are. polkadot-node does not allow `AssetTip`, while the `PolkadotExtrinsicParams` do allow this. This is why I chose base, not because it is extensible, but because it is the generic version that can be concretized to polkadot/kusmama/substrate_default or statemine/statemint etc.

So from my point of view we can either name it: `ExtrinsicParams` (then we need to rename the the ExtrinsicParams trait to ExtrinsicParamsTrait), or we call it `GenericExtrinsicParams`.

haerdib:
Renamed to Generic - I think that fits the purpose. Thanks for the input :)

Niederb (approved):
LGTM

PR #419, opened 2023-01-05
Remove thiserror

### LINKED ISSUE #317: Clean up errors: use `this_error` in `no_std` or remove completely

haerdib:
- remove this error and replace with `derive_more` where possible. This should declutter some overly verbose error messages. Better for `no_std` environments.
- clean up some unused error values
- remove generic Result in ws client - not used anyway and makes it more complex than necessary
 
closes #317

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5163,6 +5163,7 @@ dependencies = [
  "ac-compose-macros",
  "ac-node-api",
  "ac-primitives",
+ "derive_more",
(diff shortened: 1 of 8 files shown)


Niederb (approved):
Nice cleanup 👍  I like the cleanup for the Errors and that they now have more specific names.
I also learned about `derive_more` which makes this PR even better 😸

PR #421, opened 2023-01-05
Enforce `UncheckedExtrinsicV4` for `submit_extrinsic` and add `_opaque_extrinsic`

### LINKED ISSUE #90: use Bytes type in calls that send transactions

haerdib:
Renames current extrinsic submission to `opaque` and add new, more restrictive `submit_extrinisc` and similiar functions instead. This way, the encoding does not have to be implemented on client side and is less error prone as UncheckedExtrinsicV4 type is enforced.

- [x] validation against metadata? See https://github.com/scs/substrate-api-client/issues/90#issuecomment-1372355870

closes #90

### DIFF
--- a/examples/examples/batch_payout.rs
+++ b/examples/examples/batch_payout.rs
@@ -99,7 +99,7 @@ async fn main() {
 		if payout_calls_len > 0 {
 			let batching = api.batch(payout_calls);
 			let results_hash = api
-				.submit_and_watch_extrinsic_until(batching.encode(), XtStatus::InBlock)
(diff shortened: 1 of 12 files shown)


clangenb (approved):
Great, I approve!

Niederb (approved):
LGTM

PR #423, opened 2023-01-06
Enforce `Bytes` and remove obsolete hex-utils

### LINKED ISSUE #395: Investigate use of `array-bytes` or `Bytes` instead of self implemented `from_hex`
https://github.com/hack-ink/array-bytes/blob/main/src/lib.rs should do the job I think.

haerdib:
- The opaque functions now enforce `Bytes` instead of `Vec<u8>`. `Bytes` implements serialization by default with `hex_encode`, rendering the hex utils as obsolete.
- Removed utils and moved the only, not obsolete function `storage_key(..)` to `lib.rs`.

- [x] merge PR #421 

closes #395


src/lib.rs
```diff
@@ -23,9 +23,18 @@ pub use ac_compose_macros::*;
 pub use ac_node_api::*;
 pub use ac_primitives::*;
 pub use api::*;
-pub use utils::*;
 
 pub mod api;
 pub mod extrinsic;
 pub mod rpc;
-pub mod utils;
```
Niederb, line 31: I'm wondering if the tag `E1-breaksnothing` is accurate given that we remove the `utils` module. Couldn't this affect some users?

Niederb (commented):
Other than the question regarding `E1-breaksnothing` this looks good to me.

src/lib.rs
haerdib: Agreed, it does break the api. Will correct this

haerdib: Thanks for pointing it out!

Niederb (approved):

PR #424, opened 2023-01-06
Clean up Api Error and indicate future status clearly

haerdib:
- removes unused Api Error
- use clearer naming for the Api Error
- UnexptedTxStatus Error contains a clearly defined Enum, not a String. This should be easier to check for a specific status (e.g. Future), than with the previously used string.
- Removed `FinalityTimeout` from expected. Reasoning: If one waits for `Finalized`, but a Timeout is encountered, the loop would go on forever.

### DIFF
--- a/node-api/src/lib.rs
+++ b/node-api/src/lib.rs
@@ -21,7 +21,6 @@ use alloc::{borrow::ToOwned, vec::Vec};
 use codec::{Decode, Encode};
 
 pub use decoder::*;
-pub use error::*;
 pub use events::*;
 pub use metadata::*;
 pub use storage::*;
--- a/src/api/error.rs
+++ b/src/api/error.rs
@@ -15,30 +15,45 @@
 
 */
 
-use crate::{api::XtStatus, rpc::Error as RpcClientError};
+use crate::{api::UnexpectedTxStatus, rpc::Error as RpcClientError};
 use ac_node_api::{
+	error::DispatchError,
 	metadata::{InvalidMetadataError, MetadataError},
-	DispatchError,
 };
-use alloc::{boxed::Box, string::String};
+use alloc::boxed::Box;
 
 pub type Result<T> = core::result::Result<T, Error>;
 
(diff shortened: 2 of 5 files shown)


node-api/src/lib.rs
```diff
@@ -21,7 +21,6 @@ use alloc::{borrow::ToOwned, vec::Vec};
 use codec::{Decode, Encode};
 
 pub use decoder::*;
-pub use error::*;
```
haerdib, line 24: Needed to remove this one because in the api the node-api is also rexported with pub use ..::*; The default error should be the one from api-client, not the node-api.

src/api/error.rs
```diff
@@ -32,13 +32,13 @@ pub enum Error {
 	Metadata(MetadataError),
 	InvalidMetadata(InvalidMetadataError),
 	NodeApi(ac_node_api::error::Error),
-	StorageValueDecode(codec::Error),
```
haerdib, line 35: Renamed to Codec, as it's already used for general decoding errors, not just StorageValue.

src/api/error.rs
```diff
@@ -32,13 +32,13 @@ pub enum Error {
...
-	UnsupportedXtStatus(XtStatus),
```
haerdib, line 36: Removed, as it was not used at all.

Niederb (approved):
LGTM. I like the improved descriptions and the new `UnexpectedTxStatus`

node-api/src/lib.rs
clangenb: I see, in general I think we should be more conscious about exports anyhow than just wildcard re-exporting anything.

clangenb (commented):
Thanks, looks good to me!

node-api/src/lib.rs
haerdib: https://github.com/scs/substrate-api-client/issues/430 :) Thanks for the input

PR #426, opened 2023-01-06
Redo wait for events

### LINKED ISSUE #413: Investigate usage of `wait_for_event` and `wait_for_event_details`
There are two wait for event functions:
https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/src/api/rpc_api/subscribe_events.rs#L50-L92
)
Questions to be answered: 
- are still both functions needed ?
- Would it make sense so combine `wait_for_events` with the function that subscribes, instead of offering two different functions?
E.g. do not enforce two function calls such as here: https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/examples/examples/event_error_details.rs#L68-L69 The idea would be to keep it as generic as it is now. A `wait_for_success` function already exists (see [`fn submit_and_watch_extrinsic_until_success`](https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/src/api/rpc_api/author.rs#L167-L184))

subxt is generally a good start for inspiration. But be careful: subxt is fairly complex - one of the strong points of the api client for its users was its simplicity. Lets keep that as much as sensible

### LINKED ISSUE #414: Introduce `fetch_events_for_extrinsic` interface function
With the introduction of https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/src/api/rpc_api/author.rs#L167-L185
the groundwork has been laid for fetching specific events of an extrinsic. Since the logic is there, it would make sense to offer an associated interface function where the events of an extrinsic (and block? ) can be fetched.

haerdib:
@Niederb  the `subscribe_for_event_type()` function should be interesting for your runtime upgrade function.

- [ ] Clean up PR -> split up?
- [ ] Move event tests - separate file?
- [ ] add test for fn fetch_events_for_extrinsic()

closes #413 
closes #414


haerdib:
Closing in favour of separate, incoming PRs (see #428 for example)

PR #428, opened 2023-01-09
add `submit_and_watch_until_success` to examples

### LINKED ISSUE #420: Add `submit_and_watch_extrinsic_until_success` to an example
I think that's important enough of a function to be added to an example.

haerdib:
The two examples;
- `contract_instantiate_with_code`
- `event_error_details`

now use `submit_and_watch_extrinsic_until_success` and use the events reported there, instead of doing a separate `wait_for_events` call.

closes #420


examples/examples/event_error_details.rs
```diff
@@ -59,16 +59,14 @@ async fn main() {
 	println!("Sending an extrinsic from Alice (Key = {}),\n\nto Bob (Key = {})\n", alice, bob);
 	println!("[+] Composed extrinsic: {:?}\n", xt);
 
-	// Send and watch extrinsic until Ready.
-	let _tx_hash = api.submit_and_watch_extrinsic_until(xt, XtStatus::Ready).unwrap();
+	// Send and watch extrinsic until InBlock.
+	let result = api.submit_and_watch_extrinsic_until_success(xt, false);
 	println!("[+] Transaction got included into the TxPool.");
 
 	// Subscribe to system events. We expect the transfer to fail as Alice wants to transfer all her balance.
```
Niederb, line 66: Update the comment. I think just removing 'Subscribe to system events.' would be enough

Niederb (commented):
I like it and the new `xy_until_success` functions seem quite handy. For me it was not quite clear that `xy_until_success` has a different semantic than `xy_until` as it also sets the `events` member in the return value. Given that the method names are already quite long I would suggest to add some documentation in the `SubmitAndWatchUntilSuccess` trait.

examples/examples/event_error_details.rs
haerdib: Good catch, thank s !

I updated the comments - is it clearer now?

Niederb (approved):
Thanks for the detailed comments 👍

PR #429, opened 2023-01-09
add `fetch_events_from_block` and `fetch_events_for_extrinsic`

### LINKED ISSUE #414: Introduce `fetch_events_for_extrinsic` interface function

haerdib:
closes #414


src/api/rpc_api/author.rs
```diff
@@ -268,249 +267,13 @@ where
...
+		// Ensure that the extrins has been successful. If not, return an error.
+		for event in extrinsic_events.clone() {
+			event.check_if_failed()?;
+		}
 		report.events = Some(extrinsic_events);
 		Ok(report)
 	}
 }
-
-impl<Signer, Client, Params, Runtime> Api<Signer, Client, Params, Runtime>
```
haerdib, line 280: Was not changed at all, simply moved to `events.rs` because it fits there more than here.

src/api/rpc_api/events.rs
```diff
@@ -0,0 +1,372 @@
...
+				} else {
+					trace!("Not the event we are looking for, skipping.")
+				}
+			}
+		}
+		Err(Error::NoStream)
+	}
+}
+
+impl<Signer, Client, Params, Runtime> Api<Signer, Client, Params, Runtime>
```
haerdib, line 148: No new logic, simply moved from `author.rs`

src/api/rpc_api/events.rs
```diff
@@ -0,0 +1,372 @@
...
+		let block_events = self.fetch_events_from_block(block_hash)?;
+		self.filter_extrinsic_events(block_events, extrinsic_index)
+	}
+}
+
+// FIXME: This should rather be implemented directly on the
+// Subscription return value, rather than the api. Or directly
+// subcribe. Should be looked at in #288
+// https://github.com/scs/substrate-api-client/issues/288#issuecomment-1346221653
+pub trait SubscribeEvents<Client, Hash>
```
haerdib, line 79: Was not changed at all, simply renamed the file from `subscribe_events` to `events`. This will be updated in a upcoming PR (see issue #413)

src/api/rpc_api/author.rs
```diff
@@ -14,15 +14,12 @@
 //! Interface to common frame system pallet information.
```
echevrier, line 15: Only for extrinsic submission now?

haerdib: Oh, good catch. That shouldn't be frame system 👍

haerdib: Updated :)

echevrier: Yes thanks

src/api/rpc_api/author.rs
```diff
@@ -301,243 +298,13 @@ where
...
 
 		let block_hash = report.block_hash.ok_or(Error::BlockHashNotFound)?;
-		let extrinsic_index =
-			self.retrieve_extrinsic_index_from_block(block_hash, report.extrinsic_hash)?;
-		let block_events = self.fetch_events_from_block(block_hash)?;
-		let extrinsic_events = self.filter_extrinsic_events(block_events, extrinsic_index)?;
+		let extrinsic_events =
+			self.fetch_events_for_extrinsic(block_hash, report.extrinsic_hash)?;
+		// Ensure that the extrins has been successful. If not, return an error.
+		for event in extrinsic_events.clone() {
```
Niederb, line 304: Can't the events also be checked by using a reference? (in order to avoid the `clone()`)

Niederb (commented):
Looks good, just wondering if we can avoid the `clone()`

src/api/rpc_api/author.rs
haerdib: Ah yes, you're absolutely right. Fixed!

Niederb (approved):
Nice 👍

PR #434, opened 2023-01-12
Add traits and remove excessive type definition in extrinsic helpers

### LINKED ISSUE #415: Investigate if type definitions in extrinsic helpers are necessary
1. Check out: is it still up to date how extrinsic are generated? Is it really necessary to define every extrinsic call function?

Example: https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/src/extrinsic/balances.rs

haerdib:
The aim of this PR is to remove the excessive type definitions for the extrinsic creation. Until now, whenever there have been changes in types (see [an example](https://github.com/scs/substrate-api-client/pull/418/files#diff-ca9aeee70ce4891124fd150395a0a1ac0510838b362f4523c7925a6d1511cd2aL64-R90)) lots of manual changes needed to be done. To solve this, several traits are introduced in this PR.

Drawback:
- Need of trait import in case extrinsic are used.

Additionally, the following was changed:
- removed `common.rs` because all the types were for a specific pallet anyway. So moved them there.
- removed some redundant naming, e.g. `BALANCES_TRANSFER` -> `TRANSFER`
- removed unnecessary trait bounds

closes #415


src/extrinsic/staking.rs
```diff
@@ -16,235 +16,224 @@
...
+	type AccountId = Runtime::AccountId;
+
+	fn staking_bond(
 		&self,
-	) -> StakingForceNewEraXt<Signer::ExtrinsicAddress, Signer::Signature, Params::SignedExtra> {
-		compose_extrinsic!(self, STAKING_MODULE, FORCE_NEW_ERA, ForceEra {})
+		controller: Self::Address,
+		value: Self::Balance,
+		payee: Self::RewardDestination,
+	) -> ExtrinsicFor<Self, BondCall<Self::Address, Self::Balance>> {
```
clangenb, line 161: I think `ExtrinsicFor` makes it unnecessary complex, since we can use generic associated types now, I suggest:

```rust
impl StakingExtrinsic ...{
       type Extrinsic<Call> = UnchecheckedExtrinsic<Address, Call, Signature, SignedExtra>,
       
       
       fn staking_bond(...) -> Self::Extrinsic<BondCall<...>>
}

```

src/extrinsic/mod.rs
```diff
@@ -17,10 +17,46 @@
...
+impl<Signer, Client, Params, Runtime> AssignExtrinsicTypes for Api<Signer, Client, Params, Runtime>
+where
+	Signer: SignExtrinsic<Runtime::AccountId>,
+	Runtime: FrameSystemConfig,
+	Params: ExtrinsicParams<Runtime::Index, Runtime::Hash>,
+{
+	type Address = Signer::ExtrinsicAddress;
+	type Signature = Signer::Signature;
+	type SignedExtra = Params::SignedExtra;
+}
```
clangenb, line 62: I think we don't need this trait if we have the Extrinsic associated types as I described above.

src/extrinsic/staking.rs
```diff
@@ -16,235 +16,224 @@
...
-const STAKING_UNBOND: &str = "unbond";
-const STAKING_REBOND: &str = "rebond";
-const STAKING_WITHDRAW_UNBONDED: &str = "withdraw_unbonded";
-const STAKING_NOMINATE: &str = "nominate";
-const STAKING_CHILL: &str = "chill";
-const STAKING_SET_CONTROLLER: &str = "set_controller";
+use ac_primitives::{CallIndex, ExtrinsicParams, RewardDestination, SignExtrinsic, StakingConfig};
+use codec::{Compact, Decode, Encode};
+
+const MODULE: &str = "Staking";
```
clangenb, line 27: I am not sure if I agree with the plain `Module` name. I would prefer to name `STAKING_MODULE`, what was your reasoning?

haerdib: Redundancy. The file is already called `staking.rs`. Why would everything need a Staking prefix?

src/extrinsic/mod.rs
haerdib: I don't see the correlation : This trait is just so that not every trait needs to redefine these three types. Nothing to do with the Extrinsic I think?

src/extrinsic/staking.rs
haerdib: Oh! Awesome. I was actually trying that at some point, but it didn't work out. That's the much better solution, thanks for that ! ❤️ 😍 ❤️

src/extrinsic/mod.rs
haerdib: Nvm - there is a correlation.

src/extrinsic/staking.rs
clangenb: Alright, I see now that it is private, so I think it is ok. However, module is just so overly generic so I am still a bit against it.

src/extrinsic/mod.rs
clangenb: I thought `Extrinsic` is mostly where we use them as the trait is called `AssignExtrinsicTypes`, but you can quite conveniently have an extrinsic associated type to assign those values.

If this this re-assignment is needed beyond the scope of extrinsics, we should add these associated types to an overarching `ApiTrait` IMO.

src/extrinsic/staking.rs
haerdib: Kay kay, I'll give in this time. As thanks for the solution above 😉

src/extrinsic/staking.rs
haerdib: 🌞  	💞 💕💗💖 🌟💫✨ 🙀  	😍  🌋💝 🎆 🎇 🎉 🎈 🎊  📣 💎 🏆 👑 🌈 🎠 🌄 💯 💮 ‼️ 💲 🌅 ⛲

src/extrinsic/mod.rs
haerdib: I removed it. If it's about the re-assignment, as you wrote, it should be done a level above.

src/extrinsic/staking.rs
haerdib: Readded the name. But for the MODULE only. For the calls, I still think that's an overkill.

src/extrinsic/staking.rs
haerdib: And it works just like that. Wonderful 🦄 🌈 🎠💯✨

clangenb (commented):
Nice, this looks great. I only have a naming suggestion, what would you think about renaming the traits like:

CreateBalancesExtrinsic -> BalancesExtrinsics (note the pural of extrinsics).

We could also drop the `create_` prefix of the methods.

src/extrinsic/contracts.rs
```diff
@@ -17,122 +17,154 @@
...
+/// Call for instantiating a contract with the code hash.
+pub type InstantiateWithHashFor<M> =
+	(CallIndex, EndowmentFor<M>, GasLimitFor<M>, HashFor<M>, DataFor<M>);
+
+/// Call for instantiating a contract with code and salt.
+pub type InstantiateWithCodeFor<M> =
+	(CallIndex, EndowmentFor<M>, GasLimitFor<M>, CodeFor<M>, DataFor<M>, SaltFor<M>);
+
+/// Call for calling a function inside a contract.
+pub type ContractCallFor<M> = (CallIndex, AddressFor<M>, ValueFor<M>, GasLimitFor<M>, DataFor<M>);
```
Niederb, line 60: I'm wondering if we couldn't also "move" these types into the trait?
Do you think it's overkill or is there a specific reason not to do it?

Niederb (commented):
Really nice PR. Makes the "Type-Zoo" easier to digest and comprehend.

src/extrinsic/contracts.rs
haerdib: Sure, that's possible as well. If it helps to understand the code, I believe that's worth it. But maybe we should do that in a different PR, as this PR already introduces a lot of changes. What do you think?

clangenb (approved):
Thanks for the fixes, it looks great to me now!

src/extrinsic/contracts.rs
Niederb: Works for me 👍

Niederb (approved):

src/extrinsic/contracts.rs
haerdib: Here you go: https://github.com/scs/substrate-api-client/issues/441 :)

PR #437, opened 2023-01-12
Remove `wait_for_event` and simplify event access in `subscribe_events`

### LINKED ISSUE #413: Investigate usage of `wait_for_event` and `wait_for_event_details`

haerdib:
Removes the obsolete `wait_for_event` functions. I don't see the use case of these functions any more.

In addition, the event subscription has been updated:
- Renames `subscribe_frame_events` to `subscribe_events` and moved to more fitting `event.rs` instead of the frame_system file.
- Instead of returning the `StorageChangeSet`, which required quite some type insight to retrieve the events from it, a struct `EventSubscription` is returned. Upon `next_event` it directly tries to retrieve the events.

closes #413

### DIFF
--- a/examples/examples/event_callback.rs
+++ b/examples/examples/event_callback.rs
@@ -13,22 +13,16 @@
 	limitations under the License.
 */
 
-//! Very simple example that shows how to subscribe to events.
+//! Example that shows how to subscribe to events and do some action
+//! upon encounterign them.
 
-use codec::Decode;
-use frame_support::dispatch::DispatchInfo;
-use kitchensink_runtime::Runtime;
 use log::debug;
 use sp_core::H256 as Hash;
-use substrate_api_client::{
-	rpc::{HandleSubscription, JsonrpseeClient},
-	Api, PlainTipExtrinsicParams, SubscribeFrameSystem,
-};
+use substrate_api_client::{rpc::JsonrpseeClient, Api, PlainTipExtrinsicParams, SubscribeEvents};
 
-// This module depends on node_runtime.
-// To avoid dependency collisions, node_runtime has been removed from the substrate-api-client library.
+// This module depends on the specific node runtime.
 // Replace this crate by your own if you run a custom substrate node to get your custom events.
-use kitchensink_runtime::RuntimeEvent;
+use kitchensink_runtime::{Runtime, RuntimeEvent};
 
 #[tokio::main]
 async fn main() {
@@ -38,26 +32,15 @@ async fn main() {
 	let client = JsonrpseeClient::with_default_url().unwrap();
 	let api = Api::<(), _, PlainTipExtrinsicParams<Runtime>, Runtime>::new(client).unwrap();
 
-	println!("Subscribe to a single event type");
-	let (sender, receiver) = channel();
-	let api2 = api.clone();
-	thread::spawn(move || {
-		api2.subscribe_for_event_type::<ExtrinsicSuccess>(sender).unwrap();
-	});
-
 	println!("Subscribe to events");
-	let mut subscription = api.subscribe_system_events().unwrap();
+	let mut subscription = api.subscribe_events().unwrap();
 
 	// Wait for event callbacks from the node, which are received via subscription.
 	for _ in 0..5 {
-		let event_bytes = subscription.next().unwrap().unwrap().changes[0].1.clone().unwrap().0;
-		let events = Vec::<frame_system::EventRecord<RuntimeEvent, Hash>>::decode(
-			&mut event_bytes.as_slice(),
-		)
-		.unwrap();
-		for evr in &events {
-			println!("decoded: {:?} {:?}", evr.phase, evr.event);
-			match &evr.event {
+		let event_records = subscription.next_event::<RuntimeEvent, Hash>().unwrap().unwrap();
+		for event_record in &event_records {
+			println!("decoded: {:?} {:?}", event_record.phase, event_record.event);
+			match &event_record.event {
 				RuntimeEvent::Balances(balances_event) => {
 					println!(">>>>>>>>>> balances event: {:?}", balances_event);
 					match &balances_event {
@@ -75,7 +58,7 @@ async fn main() {
 				RuntimeEvent::System(system_event) => {
 					println!(">>>>>>>>>> system event: {:?}", system_event);
 					match &system_event {
-						frame_system::Event::ExtrinsicSuccess { dispatch_info: DispatchInfo } => {
+						frame_system::Event::ExtrinsicSuccess { dispatch_info } => {
 							println!("DispatchInfo: {:?}", dispatch_info);
 							return
 						},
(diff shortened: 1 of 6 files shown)


src/api/rpc_api/frame_system.rs
```diff
@@ -158,29 +155,3 @@ where
 		Ok(res)
 	}
 }
-
-pub trait SubscribeFrameSystem<Client, Hash>
```
haerdib, line 162: Moved to `event.rs` and renamed accordingly

testing/examples/frame_system_tests.rs
```diff
@@ -51,33 +51,26 @@ async fn main() {
...
+	println!("System health: {system_health}");
 
 	let system_local_peer_id = api.get_system_local_peer_id().unwrap();
-	println!("System local peer id: {:?}", system_local_peer_id);
+	println!("System local peer id: {system_local_peer_id:?}");
 
 	let system_local_listen_addresses = api.get_system_local_listen_addresses().unwrap();
-	println!("System local listen addresses: {:?}", system_local_listen_addresses);
-
-	// Subscribe
```
haerdib, line 77: Moved to events_tests.rs

testing/examples/frame_system_tests.rs
```diff
@@ -51,33 +51,26 @@ async fn main() {
 
 	// System Api
 	let system_name = api.get_system_name().unwrap();
-	println!("System name: {}", system_name);
```
haerdib, line 54: Removed some clippy warnings. Was veeery yellow here.

src/api/rpc_api/events.rs
```diff
@@ -70,24 +72,65 @@ where
...
+		EventSubscription::new(subscription)
+	}
+}
+
 pub trait SubscribeEvents<Client, Hash>
 where
 	Client: Subscribe,
 	Hash: DeserializeOwned,
 {
-	fn wait_for_event<Ev: StaticEvent>(
```
haerdib, line 82: Removed completely. It's overly complex, and with `fetch_events_for ...`, resp. the simplified `subscribe_events` I don't see where this could be useful.

examples/examples/event_callback.rs
```diff
@@ -64,8 +55,26 @@ async fn main() {
...
+				_ => debug!("ignoring unsupported module event: {:?}", event_record.event),
 			}
 		}
 	}
+
+	// After we finished whatever we wanted, unusubscribe from the subscription,
+	// to ensure, that the node does not keep sending us events.
+	// This is not mandatory. If the subscription goes out of scope, it should
+	// automatically unsubscribe from  the node.
+	subscription.unsubscribe().unwrap();
```
Niederb, line 79: I don't quite see/understand how the automatic `unsubscribe()` works. Don't we need to implement `Drop` for that to work?

clangenb: This is true, but maybe there exists a blanket implementation for that? I could imagine so, but it should be confirmed that this is the case.

testing/examples/frame_system_tests.rs
```diff
@@ -51,33 +51,26 @@ async fn main() {
...
 	let system_local_listen_addresses = api.get_system_local_listen_addresses().unwrap();
-	println!("System local listen addresses: {:?}", system_local_listen_addresses);
-
-	// Subscribe
-	let mut event_subscription = api.subscribe_system_events().unwrap();
-	let _event: ExtrinsicSuccess = api.wait_for_event(&mut event_subscription).unwrap();
-	let _event_details =
-		api.wait_for_event_details::<ExtrinsicSuccess>(&mut event_subscription).unwrap();
-	println!("Success: Wait for event Details");
+	println!("System local listen addresses: {system_local_listen_addresses:?}");
```
clangenb, line 75: I wow, this is new syntax!!

examples/examples/event_callback.rs
haerdib: Not sure what you're asking. So as the subscritpion is from the rpc-client, we basically have three different (un)subscription types. Sidenote: sadly, `drop()`  can not call `unsubscribe` due to unsubscribe requiring `self` and drop only having a reference.

For our two basic clients, the subscription is a simple channel with a receiver and waiting for the sender to send something. 

- Tungstenite: No real unsubscription anyway:  
https://github.com/scs/substrate-api-client/blob/74993c6567a74e329ae3de1dbfd7c2c9cb1636f8/src/rpc/tungstenite_client/subscription.rs#L47-L52

- ws-client:  unsubscribe and drop: https://github.com/scs/substrate-api-client/blob/74993c6567a74e329ae3de1dbfd7c2c9cb1636f8/src/rpc/ws_client/subscription.rs#L50-L62.

- jsonrpsee, for which the whole subscription trait has been designed:  handles everything internally anyway: https://github.com/scs/substrate-api-client/blob/74993c6567a74e329ae3de1dbfd7c2c9cb1636f8/src/rpc/jsonrpsee_client/subscription.rs#L32-L34 ,
resp: https://github.com/paritytech/jsonrpsee/blob/master/core/src/client/mod.rs#L218-L232


Does that answer the questions?

haerdib:
> One thing that is not quite clear to me is how the EventDetails and EventRecord relate as they seem quite similar. Is it a possibility to get rid of EventDetails?

True enough. That might be something one could look into. `EventDetails` has been copy pasted from subxt. It's basically a helper / wrapper struct. `EventRecord` is the Substrate type. But that should be investigated in a different issue. 

`EventDetails` does contain quite some convenience features. But maybe it's an overkill. Will create an issue for that.

examples/examples/event_callback.rs
clangenb: Yes, but this basically means that the comment is wrong:

``
	// This is not mandatory. If the subscription goes out of scope, it should
	// automatically unsubscribe from the node.
```

haerdib: Right. I'll correct that.

haerdib: Removed it - better now?

clangenb: Beautiful.

clangenb (approved):
Thanks a lot, lgtm now!

PR #439, opened 2023-01-13
Add `no_std` rpc client to readme

### LINKED ISSUE #393: Add explanatory no_std usage to README
After solving #279, it should be explained how to use the api client in no_std mode. E.g. 
- how to implement an rpc-client (what interfaces to implement)
- how to use Runtime in no-std (bit problematic, see  #391)

maybe a new readme needs to be added.

haerdib:
closes #393


echevrier (approved):
Thank you very much. This is really useful

Niederb (approved):
Nice documentation! LGTM

Issue #412, opened 2023-01-04
Clean up `GetStorage`


haerdib:
The current `GetStorage` ( https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/src/api/rpc_api/state.rs) uses some outdated namings.
E.g. https://github.com/scs/substrate-api-client/blob/7e5962bd2c9149f990b34aa25ad7bc7f6f92c868/src/api/rpc_api/state.rs#L59-L63
does not get the storage by Key Hash but by `StorageKey`. That should be clarified.

Additionally, some explanatory comments would be nice. Check out Substrate's comments for reference: https://github.com/paritytech/substrate/blob/7d233c2446b5a60662400a0a4bcfb78bb3b79ff7/client/rpc-api/src/state/mod.rs

PR #448, opened 2023-02-08
Sp/412 cleanup get storage

masapr:
@haerdib can you check if these are the renamings you had in mind in [#412](https://github.com/scs/substrate-api-client/issues/412)?

### DIFF
--- a/node-api/src/metadata.rs
+++ b/node-api/src/metadata.rs
@@ -455,50 +455,51 @@ impl TryFrom<RuntimeMetadataPrefixed> for Metadata {
 impl Metadata {
 	pub fn storage_value_key(
 		&self,
-		storage_prefix: &'static str,
-		storage_key_name: &'static str,
+		pallet: &'static str,
+		key: &'static str,
 	) -> Result<StorageKey, MetadataError> {
 		Ok(self
-			.pallet(storage_prefix)?
-			.storage(storage_key_name)?
-			.get_value(storage_prefix)?
+			.pallet(pallet)?
+			.storage(key)?
+			.get_value(pallet)?
 			.key())
 	}
 
 	pub fn storage_map_key<K: Encode>(
 		&self,
-		storage_prefix: &'static str,
-		storage_key_name: &'static str,
+		pallet: &'static str,
+		key: &'static str,
 		map_key: K,
 	) -> Result<StorageKey, MetadataError> {
 		Ok(self
-			.pallet(storage_prefix)?
-			.storage(storage_key_name)?
-			.get_map::<K>(storage_prefix)?
+			.pallet(pallet)?
+			.storage(key)?
+			.get_map::<K>(pallet)?
 			.key(map_key))
 	}
 
 	pub fn storage_map_key_prefix(
 		&self,
-		storage_prefix: &'static str,
-		storage_key_name: &'static str,
+		pallet: &'static str,
+		key: &'static str,
 	) -> Result<StorageKey, MetadataError> {
-		self.pallet(storage_prefix)?
-			.storage(storage_key_name)?
-			.get_map_prefix(storage_prefix)
+		self.pallet(pallet)?
+			.storage(key)?
+			.get_map_prefix(pallet)
 	}
 
+	/// `first` and `second` are the keys for the double map.
 	pub fn storage_double_map_key<K: Encode, Q: Encode>(
 		&self,
-		storage_prefix: &'static str,
-		storage_key_name: &'static str,
-		first: K,
-		second: Q,
+		pallet: &'static str,
+		key: &'static str,
+		first_double_map_key: K,
+		second_double_map_key: Q,
 	) -> Result<StorageKey, MetadataError> {
 		Ok(self
-			.pallet(storage_prefix)?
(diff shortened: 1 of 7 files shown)


src/api/rpc_api/state.rs
```diff
@@ -28,37 +28,37 @@ use serde::de::DeserializeOwned;
 pub trait GetStorage<Hash> {
 	fn get_storage_value<V: Decode>(
```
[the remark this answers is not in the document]
haerdib: Also, I think this function should be renamed to `get_storage`. The rpc call itself is called `state_getStorage` and similar functions here also do not contain any `value` (E.g `get_opaque_storage_by_storage_key`)

src/api/rpc_api/state.rs
```diff
@@ -188,85 +184,85 @@ where
...
+			self.client().request("state_getStorage", rpc_params![storage_key, at_block])?;
 		Ok(storage.map(|storage_data| storage_data.0))
 	}
 
 	fn get_storage_value_proof(
 		&self,
-		storage_prefix: &'static str,
-		storage_key_name: &'static str,
+		pallet: &'static str,
+		key: &'static str,
```
haerdib, line 211:
```suggestion
		storage_item: &'static str,
```

Maybe this clarifies it a little better what the key is for? Not sure.

node-api/src/metadata.rs
```diff
@@ -455,50 +455,41 @@ impl TryFrom<RuntimeMetadataPrefixed> for Metadata {
...
+		pallet: &'static str,
+		key: &'static str,
 	) -> Result<StorageKey, MetadataError> {
-		self.pallet(storage_prefix)?
-			.storage(storage_key_name)?
-			.get_map_prefix(storage_prefix)
+		self.pallet(pallet)?.storage(key)?.get_map_prefix(pallet)
 	}
 
+	/// `first` and `second` are the keys for the double map.
```
haerdib, line 481: Obsolete comment

src/api/rpc_api/state.rs
```diff
@@ -28,37 +28,37 @@ use serde::de::DeserializeOwned;
...
-		second: Q,
+		pallet: &'static str,
+		key: &'static str,
+		first_double_map_key: K,
+		second_double_map_key: Q,
 		at_block: Option<Hash>,
 	) -> Result<Option<V>>;
 
-	fn get_storage_by_key_hash<V: Decode>(
+	fn get_storage_by_storage_key<V: Decode>(
```
haerdib, line 59: Not sure if that's too verbose now. Maybe `get_storage_by_key` would suffice. But I'll leave the deicision to you, I don't have a hard opinion here.

src/api/rpc_api/state.rs
masapr:
```suggestion
		storage_name: &'static str,
```
Maybe we should do the naming like [here in subxt](https://github.com/paritytech/subxt/blob/92ace0629a63fadeebd428c5b1356ba60eeabfeb/metadata/src/lib.rs#L249)? So `pallet_name` and `storage_name`?

src/api/rpc_api/state.rs
```diff
@@ -121,65 +150,63 @@ where
...
-			self.metadata()
-				.storage_map_key::<K>(storage_prefix, storage_key_name, map_key)?;
+		let storagekey = self.metadata().storage_map_key::<K>(pallet, storage_item, map_key)?;
 		info!("storage key is: 0x{}", hex::encode(&storagekey));
-		self.get_storage_by_key_hash(storagekey, at_block)
+		self.get_storage_by_key(storagekey, at_block)
 	}
 
-	fn get_storage_map_key_prefix(
+	fn get_storage_key_prefix(
```
haerdib, line 176: Are you sure it's not a map prefix? To my understanding, if there's no map then the storage key can simply be calculated by combining the hashed `pallet` and `storage_item`, no need for another item.

src/api/rpc_api/state.rs
```diff
@@ -73,36 +87,51 @@ pub trait GetStorage<Hash> {
 		at_block: Option<Hash>,
 	) -> Result<Vec<StorageKey>>;
 
-	fn get_opaque_storage_by_key_hash(
+	/// Retrieve the raw storage for the given `storage_key`.
+	///
+	/// `at_block`: the state is queried at this block, set to `None` to get the state from the latest known block.
+	fn get_opaque_storage(
```
haerdib, line 93: Above, the `_by_key` is used whenever a storage key is required as an input. If there's no `by_key`, then the pallet and storage item are required. Is this convention break here intentional?

haerdib (commented):
Looking forward to the clearer interface ! Thanks for your effort ❤️

src/api/rpc_api/state.rs
masapr: changed it back

src/api/rpc_api/state.rs
masapr: Thanks for this observation. Will change it.

haerdib (approved):
Awesome, thanks alot!

PR #464, opened 2023-02-15
Remove unused `error-in-core`

haerdib:


haerdib:
Still needed it seems

Issue #460, opened 2023-02-13
Introduce empty account information fetch test


haerdib:
Currently, the api client only tests account information fetching for existing accounts: 
https://github.com/scs/substrate-api-client/blob/98aa5800bbe66bfe5ada35cec303cbe29a57760f/testing/examples/frame_system_tests.rs#L48-L50 

What is not tested, however, is if an inexistent account is queried. The expected response is `Ok(None)`.
In prior versions, an empty account fetch returned an Error, which is not okay.

PR #470, opened 2023-03-06
Test fetch information of an empty account

echevrier:
- add test in example frame_system_tests
- use Ferdie as empty account
- fix https://github.com/scs/substrate-api-client/issues/460


testing/examples/frame_system_tests.rs
```diff
@@ -48,6 +48,10 @@ async fn main() {
 	// GetAccountInformation
 	let _account_info = api.get_account_info(&alice).unwrap().unwrap();
 	let _account_data = api.get_account_data(&alice).unwrap().unwrap();
+	// Empty account information
+	let ferdie = AccountKeyring::Two.to_account_id();
+	assert!(api.get_account_info(&ferdie).unwrap().is_none());
```
haerdib, line 53: If you're doing this in one line, it might be confusing to discern what has failed (respective on which line).
It think is more intuitive if you make a two liner out of this:
```rust
let maybe_account_info = get_account_info(&ferdie).unwrap();
assert!(maybe_account_info.is_none())
```

testing/examples/frame_system_tests.rs
```diff
@@ -48,6 +48,10 @@ async fn main() {
...
+	assert!(api.get_account_data(&ferdie).unwrap().is_none());
```
haerdib, line 54:
```suggestion
```

haerdib (changes requested):

testing/examples/frame_system_tests.rs
echevrier: Don't we want to test it?

haerdib: Actually, you're right. I was thinking that account_data call account info anyway, but there's a `map`.. and we don't want that to change anything.

testing/examples/frame_system_tests.rs
echevrier: What about `let _account_data = api.get_account_data(&alice).unwrap().unwrap();` ? also put on 2 lines to know which unwrap failed?

haerdib: would be better yeah, but then we'd have to make a whole lot changes... (on the other files) and - so no, leave as is.

haerdib (approved):
Perfect, thanks !

Issue #465, opened 2023-02-15
Remove <Hash> from Balance Trait of api_rpc


haerdib:
Not necessary to be generic here: https://github.com/scs/substrate-api-client/blob/98aa5800bbe66bfe5ada35cec303cbe29a57760f/src/api/rpc_api/pallet_balances.rs#L21

Issue #455, opened 2023-02-10
impl From<Vec<u8>> for StorageKey


clangenb:
Storage Key can be found here: https://github.com/scs/substrate-api-client/blob/98aa5800bbe66bfe5ada35cec303cbe29a57760f/primitives/src/serde_impls.rs#L78

Example implementation:

https://github.com/scs/substrate-api-client/blob/98aa5800bbe66bfe5ada35cec303cbe29a57760f/primitives/src/serde_impls.rs#L37-L41

PR #471, opened 2023-03-06
impl From<Vec<u8>> for StorageKey and remove <Hash> from Balance Trait

echevrier:
- impl From<Vec<u8>> for StorageKey: https://github.com/scs/substrate-api-client/issues/455
- Remove <Hash> from Balance Trait: https://github.com/scs/substrate-api-client/issues/465

### DIFF
--- a/primitives/src/serde_impls.rs
+++ b/primitives/src/serde_impls.rs
@@ -77,6 +77,12 @@ impl From<sp_core::Bytes> for Bytes {
 )]
 pub struct StorageKey(#[serde(with = "impl_serde::serialize")] pub Vec<u8>);
 
+impl From<Vec<u8>> for StorageKey {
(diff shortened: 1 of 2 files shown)


haerdib (commented):
Looks good, thanks !

Only some minor comments:
- could you adapt the PR title to describe what you're fixing instead of linking issues?
- could you link the issues that are being closed by this PR?
- The label `E0-silent` should only be used for things that do not change functional code. Things like updating documentation, CI, and similar. For code changes that do not break anything, please use the label `E1-breaksnothing`.
The reason is: E0 labels will not be added in the automatic release text, which describes what has changed compared to the previous releae. Whereas E1 >= labels will be added.

haerdib (approved):
Awesome, thanks !

PR #478, opened 2023-03-22
Experiment with removing the u64 implementations in order to support 32 bit architectures

Niederb:


Niederb:
@haerdib Can you have a look? I would rebase once master is working again and then we can merge it.

echevrier (approved):
LGTM

Issue #474, opened 2023-03-13
Fix security issue of dependabot


haerdib:
See https://github.com/scs/substrate-api-client/security/dependabot/10

PR #479, opened 2023-03-22
update to newest substrate

echevrier:
- Upgrade to substrate commit : https://github.com/paritytech/substrate/tree/8bd57418b79d2450079755c0d631168b8fad2e97
    - Change Keystore according to https://github.com/paritytech/substrate/commit/bf395c8308c481a9774373e0b0b14bd7a2e4b8d2 and https://github.com/paritytech/substrate/commit/9d2d020c5134d8c7870396cb04bca63a674abfd5
    - Change Balances because of https://github.com/paritytech/substrate/pull/12951. Note that
          - free, locked and reserved balances are defined differently  
          - Existential Deposit requirements to be made of the free balance, not the total balance
          - set_balance is new force_set_balance
          - transfer is transfer_allow_death
- Cargo update
- Fix the tests but the [DispatchError](https://github.com/scs/substrate-api-client/blob/33622f4d3e6a949e1a6efa507cff58173d74d856/node-api/src/error.rs#L79) should be improved to differentiate all DispatchError::Other. See issue https://github.com/scs/substrate-api-client/issues/488

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,7 +249,7 @@ checksum = "86ea188f25f0255d8f92797797c97ebf5631fa88178beb1a46fdf5622c9a00e4"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.5",
+ "syn 2.0.8",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


haerdib (commented):
Cool, thanks for the fix!

Niederb (approved):
LGTM. Seems you always get the nasty upgrades 😆 
On the plus side the `cargo update` fixed #474 (`remove_dir_all` is not in the `Cargo.lock` anymore)

haerdib (approved):
🥳

Issue #504, opened 2023-03-30
`ExtrinsicSigner` struct does not allow clone


haerdib:
I've run into the problem that the ExtrinsicSigner created in the worker does not seem to be cloneable. I haven't looked further into this yet. Therefore, I see the following TODOs for this issue:

Steps:

- [x] Try to reproduce by creating an unit test and clone the `ExtrinsicSigner`
- [ ] Fix if possible & reproducible -> not possible because the Unlookup struct from substrate does not implement clone.

Workaround: Implementing a new ExtrinsicSigner that has a hardcoded Address type: see https://github.com/scs/substrate-api-client/pull/511

But before investing too much time, this issue may be obsolete after finishing #406.

PR #511, opened 2023-04-03
Add static signer

### LINKED ISSUE #461: Introduce a no-wasm compatible `ExtrinsicSigner`
The current `ExtrinsicSigner` depends on a `Runtime`. But the node Runtime is not available in an environment where wasm is not supported.

Therefore, a No wasm compatible `ExtrinsicSigner` (implementing the same trait `SignExtrinsic`) where no Runtime type input is needed could be introduced.

If not possible in any other way, a hardcoded extrinsic_address type conversion may be used.

@clangenb Would that help your usecase?

haerdib:
closes #461

workaround for #504

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,6 +86,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sp-core",
+ "sp-keyring",
  "sp-runtime",
  "sp-staking",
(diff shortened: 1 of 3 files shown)


Niederb (approved):
LGTM

Issue #318, opened 2022-11-14
Software Design: Introduce nonce caching


haerdib:
Currently, every `compose_extrinsic!` call, calls the substrate storage. There's already a workaround for this by using `ccompose_extrinsic_offline!`, but it sure would be nice, to have a nonce cache that caches the nonce, and updates it with a storage request in case of a failure (e.g. mark it as dirty or something). 

Complexity: When to revert nonce? What about future?

Niederb:
Maybe in a first version we can introduce a `NonceProvider` trait that abstracts over getting the nonce and provide two simple implementations:
- `LocalNonceProvider`: Fetches the nonce only once and increases it automatically, but does not implement additional logic (fetching nonce again on error, ...)
- `OnlineNonceProvider`: The approach currently used that always fetches the nonce from the node

This will already be some work to think about how the trait should look, how to configure this and how to pass the `NonceProvider` around. We could then implement fancier nonce caching later on.

haerdib:
I agree in general, but before doing all this work, we should probably make sure that a local nonce cache does make sense. At least theoretically. E.g. find out if such a cache could ever be robust enough to be applied in the field. What do you think?

Niederb:
Yes, I agree and I think even this "simple" approach is more involved that it seems at first glance.

PR #521, opened 2023-04-05
Experiment with async

Niederb:
Only make the `request` method async without breaking external APIs


src/api/api_client.rs
```diff
@@ -177,26 +177,32 @@ where
...
 {
 	/// Create a new Api client with call to the node to retrieve metadata.
-	pub fn new(client: Client) -> Result<Self> {
-		let genesis_hash = Self::get_genesis_hash(&client)?;
+	pub async fn new(client: Client) -> Result<Self> {
+		let genesis_hash_future = Self::get_genesis_hash(&client);
+		let metadata_future = Self::get_metadata(&client);
+		let runtime_version_future = Self::get_runtime_version(&client);
+
+		let (genesis_hash, metadata, runtime_version) = futures::future::try_join3(
```
Niederb, line 185: This unfortunately leads to problems with `no_std` 😢 
We could just make a couple of `await`s instead but it would be nice to wait in parallel.

src/api/api_client.rs
```diff
@@ -177,26 +177,32 @@ where
...
-	pub fn update_runtime(&mut self) -> Result<()> {
-		let metadata = Self::get_metadata(&self.client)?;
-		debug!("Metadata: {:?}", metadata);
+	pub async fn update_runtime(&mut self) -> Result<()> {
+		let metadata_future = Self::get_metadata(&self.client);
+		let runtime_version_future = Self::get_runtime_version(&self.client);
 
-		let runtime_version = Self::get_runtime_version(&self.client)?;
+		let (metadata, runtime_version) =
+			futures::future::try_join(metadata_future, runtime_version_future).await?;
```
Niederb, line 204: Same as above

src/lib.rs
```diff
@@ -16,6 +16,7 @@
 */
 #![cfg_attr(not(feature = "std"), no_std)]
 #![feature(error_in_core)]
+#![feature(async_fn_in_trait)]
```
Niederb, line 19: Using `async` in traits is not available on stable yet

Niederb:
@haerdib This is not finished yet, but can you have a first look?
Most of it is boilerplate of converting functions to async and using await.
Some thoughts (also see comments):
- We need a new nightly feature to have support for async in traits
- It would be nice to use features from `futures::future` but this needs `std`
- I'm not sure what the situation is for executor support in no_std. For example in wasm. I'm just wondering if we limit our potential use-cases by switching to async.

haerdib (commented):
> I'm not sure what the situation is for executor support in no_std. For example in wasm. I'm just wondering if we limit our potential use-cases by switching to async.

This would certainly need thorough investigation. But instead of switching completely to async, have you considered providing both, async and sync client? That would solve the question if it's available in wasm and such - if it's not, one can simply use the sync client.

Ofc, that would need quite some thought on how to provide two clients without having to duplicate everything.

Niederb:
- I looked a bit more into [maybe-async](https://crates.io/crates/maybe-async). It uses the `async_trait` crate to support async on traits (instead of the nightly feature). 
  - I opened an issue asking about support for nightly: https://github.com/fMeow/maybe-async-rs/issues/17
- The nightly implementation currently does not support [dynamic dispatch](https://blog.rust-lang.org/inside-rust/2022/11/17/async-fn-in-trait-nightly.html#limitation-dynamic-dispatch) which seems quite limiting to me so we might be better of using `async-trait` anyway.

compose-macros/src/lib.rs
```diff
@@ -101,7 +102,7 @@ macro_rules! compose_extrinsic {
                 $crate::compose_extrinsic_offline!(
                     signer,
                     call.clone(),
-                    $api.extrinsic_params($api.get_nonce().unwrap())
```
Niederb, line 104: `get_nonce` is now (potentially) async, but unfortunately the `maybe-async`-magic does not work inside of macros 😢 
So we need to get the nonce already outside and pass it in.

Niederb:
Still not finished but I think it is closer to what we want.
The approach is now quite different:
- Use `maybe-async` and `async-trait` crates instead of nightly implementation.
  - This allows us to provide a sync and an async API by changing a compiler flag
  - Examples compile without changes
- The only difference between async and sync is in `JsonrpseeClient::request`. There is now a separate implementation for each.
@haerdib What do you think?

Cargo.toml
```diff
@@ -61,22 +64,17 @@ disable_target_static_assertions = [
     "sp-runtime-interface/disable_target_static_assertions",
 ]
 std = [
-    # crates.io no_std
```
haerdib, line 64: Why remove these comments?

compose-macros/src/lib.rs
haerdib: This is a pity, but I never liked the `unwrap` inside the macro anyway. So I think that's an acceptable approach. But as a user-friendly macro alternative: How about offering a function called `compose_extrinisc` that does retrieve the nonce and call the macro? Just like here: 
https://github.com/scs/substrate-api-client/blob/6d09a9de7ce0fe7848d1bf2c206259bb1eceaea3/src/extrinsic/offline_extrinsic.rs#L32-L43

haerdib (commented):
> - Examples compile without changes

Does that mean that this PR does not introduce any breaking changes? E.g. the users do not need to add the `#[maybe_async::maybe_async(?Send)]` in case they use the good old sync variant?

compose-macros/src/lib.rs
Niederb: The problem with the `compose_extrinisc` macro is that it takes a variable number of arguments, so it cannot as easily be wrapped into a function.

haerdib: Right, I forgot about that part. It's fine like this, but just as another idea: implementing a strictly sync version for get_nonce is not possible in this setup?

Niederb: It certainly would be nice but since we get the nonce from a node, I think it should be async. Otherwise our `async` support is quite limited when a big chunk of our requests are still sync. 
Maybe we can think of something in #318 but even there we might have to fetch the nonce from a node and then it is still `async`.
Now that I think about it a bit more.... the nonce is quite limiting for concurrency/async in general. In an async use case with a single account the likelihood of having multiple transactions with the same nonce is probably quite high. This is kind of a by design of polkadot though.

haerdib: What's the use case for an async nonce call in extrinsic creation? You have to know the nonce, otherwise you can't create the extrinsic. 

Yes, #318 would provide a different solution approach. The nonce cache would also solve your concern about the async use case. The extrinsic creation is not time-consuming if the nonce is provided locally. So no need to make it async, I believe. Only the extrinsic sending should be made available as async. If the node allows Future, the order of sending does not matter.

Niederb:
I think this is now ready for merging. The nonce handling is now the same as before. 
I created a follow up task for implementing an example #543

Cargo.toml
```diff
@@ -24,11 +24,13 @@ members = [
 
 [dependencies]
 # crates.io no_std
+async-trait = "0.1.68"
```
haerdib, line 27: Still needed?

Cargo.toml
```diff
@@ -54,17 +56,20 @@ ac-compose-macros = { path = "compose-macros", default-features = false }
...
 scale-info = { version = "2.1.1", features = ["derive"] }
 
 [features]
-default = ["std", "jsonrpsee-client"]
+default = ["std", "jsonrpsee-client", "sync-api"]
 # To support `no_std` builds in non-32 bit environments.
 disable_target_static_assertions = [
     "sp-runtime-interface/disable_target_static_assertions",
 ]
+# If this is active all the code compiles in synchronous mode. Otherwise `async` is supported
```
haerdib, line 71:
```suggestion
# If this is active all the code compiles in synchronous mode. Otherwise `async` is supported.
```

compose-macros/src/lib.rs
```diff
@@ -29,8 +29,8 @@ mod rpc;
 /// # Arguments
 ///
 /// * 'node_metadata' - This crate's parsed node metadata as field of the API.
-/// * 'module' - Module name as &str for which the call is composed.
-/// * 'call' - Call name as &str
+/// * 'pallet' - Module name as &str for which the call is composed.
```
haerdib, line 32:
```suggestion
/// * 'pallet' - Pallet name as &str for which the call is composed.
```

compose-macros/src/lib.rs
```diff
@@ -78,31 +75,70 @@ macro_rules! compose_extrinsic_offline {
...
-macro_rules! compose_extrinsic {
+macro_rules! compose_extrinsic_with_nonce {
 	($api: expr,
+	$nonce: expr,
 	$module: expr,
 	$call: expr
 	$(, $args: expr) *) => {
 		{
-            use $crate::log::debug;
+			use $crate::log::debug;
```
haerdib, line 91: Is that indent correct?

compose-macros/src/lib.rs
```diff
@@ -78,31 +75,70 @@ macro_rules! compose_extrinsic_offline {
 /// # Arguments
 ///
 /// * 'api' - This instance of API. If the *signer* field is not set, an unsigned extrinsic will be generated.
+/// * 'nonce' - signer's account nonce: u32
```
haerdib, line 78:
```suggestion
/// * 'nonce' - signer's account nonce: Index
```

flipchan:
@Niederb awesome new feature! is this stable?

src/api/api_client.rs
haerdib: I think that's fine. The async improvement is worth the small duplication.

haerdib (commented):
Absolutely awesome, thanks a lot!

One thing: Could you rename the PR to a little more meaningful name and link the corresponding issue (though not close it I guess because Subscription is not yet async)

compose-macros/src/lib.rs
Niederb: No, I fixed it now. Seems `cargo fmt` works differently in macros. Interesting

Niederb:
> @Niederb awesome new feature! is this stable?

@flipchan Thanks! I'm not quite sure what you mean by stable? I hope this PR will be merged very soon and then async will be available for the more simple communication that uses a single request/response (more complex communication based on subscriptions is not done yet). 
`substrate-api-client` needs a nightly rust compiler because of other features we need.

Niederb:
> Absolutely awesome, thanks a lot!
> 
> One thing: Could you rename the PR to a little more meaningful name and link the corresponding issue (though not close it I guess because Subscription is not yet async)

Done, but linking without closing is apparently not supported. See https://github.com/orgs/community/discussions/17308
I guess we can just repopen the issue after the merge 😉

Cargo.toml
haerdib: Interesting. Thanks for the explanation!

haerdib (approved):
👏

echevrier (approved):
Thanks a lot. Really nice!

PR #522, opened 2023-04-06
Make `pallet-staking` and `pallet-contracts` dependencies optional

Niederb:
Include `pallet-staking` dependency only when staking-xt feature is enabled.
`pallet-contracts` is implemented along the same lines. `pallet-contracts` has a dependency on `wasmi` so it is nice if we can make this optional.


clangenb (approved):
Looks great to me!

haerdib (approved):
Cool, thanks !

PR #532, opened 2023-04-13
Clarify distinction of `Metadata` struct and internal `metadata` field

### LINKED ISSUE #493: Make Metadata struct more intuitive
The current Metadata struct contains a field metadata, but also some other stuff. That's absolutely great, the only problem: The difference between the runtime metadata and the Metadata struct is not clear, as both are called metadata. Maybe the struct should be renamed. Or the "raw" metadata access should be made more intuitive.
E.g. serializing should automatically only serialize the raw metadata.

haerdib:
To make it more obvious to the user what's the difference between `Metadata` and `metadata` the following is updated:
- renamed `metadata` to `runtime_metadata`
- Added explanatory comment to `Metadata`struct
- `pretty_format` now takes self as argument and returns an error instead of an Option (which resulted in information loss in case of an error)

closes #493

### DIFF
--- a/examples/examples/print_metadata.rs
+++ b/examples/examples/print_metadata.rs
@@ -18,9 +18,7 @@
 
 use kitchensink_runtime::Runtime;
 use sp_core::sr25519;
-use substrate_api_client::{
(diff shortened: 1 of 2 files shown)


echevrier (approved):
Nice. Small but really useful changes

Niederb (approved):
Great 🎈

PR #535, opened 2023-04-13
Clean up dependencies and fix primitives no_std build

haerdib:
As preparatory steps for publishing on crates.io:

1. Clean up the toml files:
    - remove unused dependencies
    - import frame-metadata from crates.io instead of github (less changes necessary for publishing on crates.io)
    - complete the toml file description of each crate

2. Additionally, found while cleaning dependencies:
    - fix primitives crate no_std build and include the build test in CI

3. Removed `assert` in error example because it constantly failed (substrate upgrade?). But since there is no decoding anyway, I don't see the use case in checking the encoded error.

4. Last, but not least:
    - add READMEs to each crate


primitives/Cargo.toml
```diff
@@ -4,22 +4,29 @@ version = "0.6.0"
...
 
 # substrate no_std
-sp-core = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-core = { default-features = false, features = ["full_crypto"], git = "https://github.com/paritytech/substrate.git", branch = "master" }
 sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-runtime-interface = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
 sp-staking = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
 sp-version = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
 sp-weights = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+# need to add this for the app_crypto macro
```
Niederb, line 28: This is not quite clear to me. Can you provide a bit more context?

primitives/Cargo.toml
```diff
@@ -4,22 +4,29 @@ version = "0.6.0"
...
-hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
 impl-serde = { version = "0.4", default-features = false }
 primitive-types = { version = "0.12.1", default-features = false, features = ["serde_no_std", "scale-info"] }
 scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
 serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 
 # substrate no_std
-sp-core = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "master" }
+sp-core = { default-features = false, features = ["full_crypto"], git = "https://github.com/paritytech/substrate.git", branch = "master" }
```
Niederb, line 22: I assume these changes are needed to get `no_std` working?

haerdib: Yes, exactly. The trait `Pair` 
https://github.com/scs/substrate-api-client/blob/6d09a9de7ce0fe7848d1bf2c206259bb1eceaea3/primitives/src/extrinsics/signer.rs#L23

is only available in std or `full_crypto`. The reason it compiled without problems until now, is that the `full_crypto` feature has been added by the importing crates.

primitives/Cargo.toml
haerdib: The problem is that `sp-application-crypto` is indirectly imported by some other crate of substrate, but without the feature `full_crypto`. To get it working, however, we need this feature, so we import it directly to activate it.

I haven't looked too deeply into what this app_crypto thing is, but there is a `AppCrypto` trait here: https://github.com/paritytech/substrate/blob/master/primitives/application-crypto/src/traits.rs#L34
And it is a fact that it does not compile without activating the full_crypto feature.

Niederb (approved):
Nice cleanup and improvement of no_std verification 🥳

PR #539, opened 2023-04-18
Add a new function storage_double_map_key_prefix to meta to implement…

bai-3:
use as this
```
let key = api
    .get_storage_double_map_key_prefix("WeteeGov", "ReferendumInfoOf", 5000)
    .unwrap();

let storage_keys = api
    .get_storage_keys_paged(Some(key), 10, None, None)
    .unwrap();

for storage_key in storage_keys.iter() {
    println!("Retrieving value for key {:?}", storage_key);
    let storage_data: ReferendumInfo<BlockNumber, RuntimeCall, Balance> = api
        .get_storage_by_key_hash(storage_key.clone(), None)
        .unwrap()
        .unwrap();
    println!("Retrieved data {:?}", storage_data);
}
```

### DIFF
--- a/src/api/rpc_api/state.rs
+++ b/src/api/rpc_api/state.rs
@@ -59,6 +59,13 @@ pub trait GetStorage {
 		storage_item: &'static str,
 	) -> Result<StorageKey>;
 
+	fn get_storage_double_map_key_prefix<K: Encode>(
+		&self,
+		storage_prefix: &'static str,
+		storage_key_name: &'static str,
+		first: K,
+	) -> Result<StorageKey>;
+
 	/// Retrieve the storage value from a double map for the given keys: `first_double_map_key` and `second_double_map_key`.
 	///
 	/// `at_block`: the state is queried at this block, set to `None` to get the state from the latest known block.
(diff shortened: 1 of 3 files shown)


src/api/rpc_api/state.rs
```diff
@@ -59,6 +59,13 @@ pub trait GetStorage {
 		storage_item: &'static str,
 	) -> Result<StorageKey>;
 
+	fn get_storage_double_map_key_prefix<K: Encode>(
+		&self,
+		storage_prefix: &'static str,
+		storage_key_name: &'static str,
+		first: K,
```
haerdib, line 66:
```suggestion
		first_key: K,
```

src/api/rpc_api/state.rs
```diff
@@ -59,6 +59,13 @@ pub trait GetStorage {
 		storage_item: &'static str,
 	) -> Result<StorageKey>;
 
+	fn get_storage_double_map_key_prefix<K: Encode>(
```
haerdib, line 62: Could you add an explanatory comment? Something like 
`/// Retrieve the key prefix for a storage double map with the first key included.`

haerdib (commented):
Thank you very much for your contribution! 

Would it be possible to add an example call to the testing folder? Just like with the `storage_map_key_prefix`:


https://github.com/scs/substrate-api-client/blob/19588b1c29a75a3defdbcc65d0bdfd796732f72b/testing/examples/state_tests.rs#L63-L67

That way, CI will test the new function automatically.

Available DoubleMaps in CI would for example be: 
- Staking pallet `EraStakers`: https://github.com/paritytech/substrate/blob/785115b3a13901b0c708af8166430bcc9c71f28f/frame/staking/src/pallet/mod.rs#L404-L415
- Assets `Account`: https://github.com/paritytech/substrate/blob/785115b3a13901b0c708af8166430bcc9c71f28f/frame/assets/src/lib.rs#L327-L334

However, these will be all empty in the CI node. But for key comparison, that should still work.

Issue #440, opened 2023-01-18
Investigate usage of `EventDetails`


haerdib:
According to : https://github.com/scs/substrate-api-client/pull/437#issuecomment-1386835667

> One thing that is not quite clear to me is how the EventDetails and EventRecord relate as they seem quite similar. Is it a possibility to get rid of EventDetails?

True enough. That might be something one could look into. `EventDetails` has been copy pasted from subxt. It's basically a helper / wrapper struct. `EventRecord` is the Substrate type. But that should be investigated in a different issue. 

`EventDetails` does contain quite some convenience features. But maybe it's an overkill. Will create an issue for that.

haerdib:
So, the main difference is that `EventRecord` needs an `RuntimeEvent` Type. Which needs to be taken directly from the Runtime, as there's some macro magic behind it, creating a Type with all the Events in it. Or something like that.

On the other hand, `EventDetails` do not need this magically created RuntimeEvent type. Instead, they need to know the Metadata. 

On one hand, the RuntimeEvent Type is absolutely beautiful: 

https://github.com/scs/substrate-api-client/blob/303ecebad02c7f551cabda9fa102e98e06300145/examples/examples/event_callback.rs#L45-L53

But, just like the `Runtime` itself, that's a little problematic for `no_std` environments. So throwing away `EventDetails` is no good IMO.

The main problem now: We use both types. It works, but I didn't give it much thought when implementing it. So the next step would be: Does it make sense to use both types as it is now, or is it some strange, random mix?

haerdib:
I think it makes sense to offer both variations. Once the self explanatory (default) type RuntimeEvents and one, when needed, the EventDetails from metadata as both have their down- but upsides.

See #545

PR #545, opened 2023-04-21
Add `next_events_from_metadata` and rename `next_event`

haerdib:
Updated the `EventSubscription` Wrapper to support decoding events from the metadata:
- added new functionality `next_events_from_metadata` to derive events from metadata instead of `RuntimeEvents` generated by the `construct_runtime` macro of Substrate
- renamed `next_event` to more appropriate `next_events`.
- introduced a `metadata` field for event derivation from metadata (needed by `next_events_from_metadata`). Therefore, the `From` derivation from `Subscription`is not possible any more (needs metadata) and was removed.

closes https://github.com/scs/substrate-api-client/issues/440

### DIFF
--- a/node-api/src/events.rs
+++ b/node-api/src/events.rs
@@ -24,9 +24,6 @@ use sp_core::H256;
 
 /// A collection of events obtained from a block, bundled with the necessary
 /// information needed to decode and iterate over them.
-//
-// In subxt, this was generic over a `Config` type, but it's sole usage was to derive the
-// hash type. We omitted this here and use the `ac_primitives::Hash` instead.
 #[derive(Clone, Debug)]
 pub struct Events<Hash> {
 	metadata: Metadata,
--- a/examples/examples/event_callback.rs
+++ b/examples/examples/event_callback.rs
@@ -39,7 +39,7 @@ async fn main() {
 
 	// Wait for event callbacks from the node, which are received via subscription.
 	for _ in 0..5 {
-		let event_records = subscription.next_event::<RuntimeEvent, Hash>().unwrap().unwrap();
+		let event_records = subscription.next_events::<RuntimeEvent, Hash>().unwrap().unwrap();
 		for event_record in &event_records {
 			println!("decoded: {:?} {:?}", event_record.phase, event_record.event);
 			match &event_record.event {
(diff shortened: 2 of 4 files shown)


Niederb (approved):
LGTM

echevrier (commented):
For me it is not clear what this PR wants to do. Can you clarify? 
Why does this PR close #440?
Do we still need `RuntimeEvent`? When should we use `next_events`, when `next_events_from_metadata`?

node-api/src/events.rs
```diff
@@ -24,9 +24,6 @@ use sp_core::H256;
 
 /// A collection of events obtained from a block, bundled with the necessary
 /// information needed to decode and iterate over them.
-//
-// In subxt, this was generic over a `Config` type, but it's sole usage was to derive the
-// hash type. We omitted this here and use the `ac_primitives::Hash` instead.
```
haerdib, line 29: Removed because outdated. There's no ac_primitves::Hash Type. No need to point out what Subxt did, because Subxt is constantly evolving as well.

haerdib:
I updated the PR description, better now?

Regarding the question when to use what: is the comment 
https://github.com/scs/substrate-api-client/blob/7536d15a57866b949de97ac2fc81f75b71c6ac0b/src/api/rpc_api/events.rs#L124-L128
not explanatory enough?

haerdib:
> Do we still need RuntimeEvent
Check out https://github.com/scs/substrate-api-client/issues/440#issuecomment-1516501287 for more elaborate explanation.
But the short answer would be: no, we don't need it. But I think it would be a waste to not offer it.

echevrier:
> I updated the PR description, better now?
> Regarding the question when to use what: is the comment
> 
> https://github.com/scs/substrate-api-client/blob/7536d15a57866b949de97ac2fc81f75b71c6ac0b/src/api/rpc_api/events.rs#L124-L128
> 
> not explanatory enough?

It suggests to use every time you can `next_events `and the `next_events_from_metadata` when you really need it. Your comment https://github.com/scs/substrate-api-client/issues/440#issuecomment-1523108879 confirms it. I would like to have this information in the PR description as we do not have examples, which explicitly show the different use cases. This would help the user, without having to click/read the PR and related issues.

echevrier (approved):

PR #552, opened 2023-04-26
Add documentation for the websocket features

### LINKED ISSUE #546: `async` support for tungstenite and ws clients
Currently only the `jsonrpsee` client has actual support for `async`, the others simply block no matter what.
If possible we should implement async for the `tungstenite` and `ws` clients as well. If not possible we should document that async is not supported.

Niederb:
See my comment in the issue: https://github.com/scs/substrate-api-client/issues/546#issuecomment-1521455679
Closes #546

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,8 +91,11 @@ std = [
     "ac-node-api/std",
     "ac-primitives/std",
 ]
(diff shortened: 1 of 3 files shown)


haerdib (approved):
Cool, thanks a lot!

echevrier (approved):

Issue #543, opened 2023-04-19
Provide an example for async requests


Niederb:
#278 introduces support for sending async requests to the node. We should have an example that shows how to use this capability.

PR #556, opened 2023-04-28
Create no_std example for extrinsic creation

Niederb:
Adjusted copy of `compose_extrinsic_offline`.
Closing #543


Niederb:
@haerdib Not finished yet but can you have a look if this is the functionality you imagined?

- Next step would be to extract the extrinsic creation into a separate function (see issue #543 )
  - Will be a bit involved because lots of types will have to be made explicit
  - It would be nice to actually compile this part in no_std mode but I don't see an easy way to do it.
- If possible merging the code into `compose_extrinsic_offline` to avoid redundancy
  - Check how easy it is once the function is ready

examples/examples/compose_extrinsic_no_std.rs
```diff
@@ -0,0 +1,95 @@
...
+		.tip(0);
+	let extrinsic_params = GenericExtrinsicParams::new(
+		spec_version,
+		transaction_version,
+		signer_nonce,
+		genesis_hash,
+		additional_extrinsic_params,
+	);
+	let call =
+		RuntimeCall::Balances(BalancesCall::transfer_allow_death { dest: recipient, value: 42 });
```
haerdib, line 88: Same here for the `RuntimeCall`: In no_std the runtime compiles to wasm, so tricky to import in no_std environments. Better to go without it. That's the fun part now: To get the Call you need the metadata.

Check out how it's done in the worker:
metadata call: https://github.com/integritee-network/worker/blob/master/core-primitives/node-api/metadata/src/lib.rs#L72-L88
Retrieve indexes for a specific pallet: https://github.com/integritee-network/worker/blob/master/core-primitives/node-api/metadata/src/pallet_teeracle.rs

haerdib (commented):
This definitely goes into the right direction. 

Regarding the comment:
> Next step would be to extract the extrinsic creation into a separate function 

While I wrote that in the issue, that's just a thought on how to make it clear what's std and not. If you have a better idea or if it's too complicated, then feel free to adapt. 

> It would be nice to actually compile this part in no_std mode but I don't see an easy way to do it.

Neither do I... move it to another crate and everything is an overkill I believe. So let's do it with best effort. That should suffice.

examples/examples/compose_extrinsic_no_std.rs
```diff
@@ -0,0 +1,105 @@
...
+	let extrinsic_signer = ExtrinsicSigner::<_>::new(signer);
+	// Signer is needed to get the nonce
+	api.set_signer(extrinsic_signer.clone());
+
+	// Information for Era for mortal transactions (online).
+	let last_finalized_header_hash = api.get_finalized_head().unwrap().unwrap();
+	let header = api.get_header(Some(last_finalized_header_hash)).unwrap().unwrap();
+	let period = 5;
+
+	// Get information out of Api (online).
```
haerdib, line 65:
```suggestion
	// Get information out of Api (online). This information could also be set offline in the `no_std`, but that would need to be static and adapted whenever the node changes. You can get the information directly from the node runtime file or the api of https://polkadot.js.org.
```

examples/examples/compose_extrinsic_no_std.rs
```diff
@@ -0,0 +1,105 @@
...
+	// because of the constantly changing substrate node. But check out our unit tests - there are Apis created with `new_offline`.
+	//
+	// ! Careful: AssetTipExtrinsicParams is used here, because the substrate kitchensink runtime uses assets as tips. But for most
+	// runtimes, the PlainTipExtrinsicParams needs to be used.
+	let mut api = Api::<SubstrateKitchensinkConfig, _>::new(client).unwrap();
+	let extrinsic_signer = ExtrinsicSigner::<_>::new(signer);
+	// Signer is needed to get the nonce
+	api.set_signer(extrinsic_signer.clone());
+
+	// Information for Era for mortal transactions (online).
```
haerdib, line 60:
```suggestion
	// Get the last finalized header to retrieve information for Era for mortal transactions (online).
```

haerdib (commented):
Cool, thanks a lot for the effort! 

Two more things:
1. Can you integrate this example into CI?
2. Any chance of merging the code into compose_extrinsic_offline to avoid redundancy, as you've mentioned?

haerdib:
Is this PR open for Review btw?

Niederb:
> Is this PR open for Review btw?

@haerdib Unfortunately still not quite... I merged the examples but I'm still not quite happy how the code looks. Also the CI integration is missing.

Niederb:
The two examples are now merged into a single one. Both approaches are in the `main()`, extracting separate functions would need more type annotations. There is also still some code duplication which is due to use of different generic types, but overall I think the code is relative straightforward.

examples/examples/compose_extrinsic.rs
```diff
@@ -0,0 +1,140 @@
...
+
+	// Construct extrinsic without using Api (no_std).
+	let additional_extrinsic_params: AdditionalParams = GenericAdditionalParams::new()
+		.era(Era::mortal(period, header.number.into()), last_finalized_header_hash)
+		.tip(0);
+
+	println!("Compose extrinsic in no_std environment (No Api instance)");
+	{
+		let signer_nonce = api.get_nonce().unwrap();
+		println!("[+] Alice's Account Nonce is {}", signer_nonce);
```
haerdib, line 79: Would it be possible to move the api part outside of this environment? That would make it clearer, that really no api client is needed (given the information is availbale).
E.g.
```rust
let spec_version = api.runtime_version().spec_version;
let transaction_version = api.runtime_version().transaction_version;
let genesis_hash = api.genesis_hash
..
let mut xt = ...;
{
        let extrinsic_params = <AssetRuntimeConfig as Config>::ExtrinsicParams::new(
			spec_version,
			transaction_version,
			signer_nonce,
			genesis_hash,
			additional_extrinsic_params,
		);

		let recipients_extrinsic_address: ExtrinsicAddressOf<AssetExtrinsicSigner> =
			recipient.clone().into();

		let call = compose_call!(
			metadata,
			"Balances",
			"transfer_allow_death",
			recipients_extrinsic_address,
			Compact(4u32)
		);
		xt = compose_extrinsic_offline!(extrinsic_signer, call, extrinsic_params);
}
// Test created extrinsic via api call to the node (online)
let hash = api
			.submit_and_watch_extrinsic_until(xt, XtStatus::InBlock)
			.unwrap()
			.block_hash
			.unwrap();

haerdib: Same for the api (offline) part - move the online part outside? 

Does that make sense?

haerdib (approved):
Cool, thanks a lot!

PR #557, opened 2023-04-28
Bump cargo-install to remove Node.js warnings

haerdib:
Fixes the Node.js warnings (see Annotations here: https://github.com/scs/substrate-api-client/actions/runs/4806291597?pr=541)


Niederb (approved):
Nice 👍

PR #558, opened 2023-04-28
Document feature flags

### LINKED ISSUE #526: Add documentation for rpc-client / features import
This should include description on:
- Difference between rpc clients (e.g. when to use which client)
    - jsonrspee is only opens one websocket connection to the node (making it most likely the most efficient /fastest client). Offers the most convenience functions (subscription callbacks and such). Depends on tokio runtime.
    - ws client: Does not offer any convenience functions, but is simple and easily adaptable. Does not depend on tokio runtime. 
    - tungstenite: offers (limited) reconnection possibilites. Other than that, similiar to ws-client.
- staking / contracts extrinsic import (see #522)

Niederb:
Closes #526


Cargo.toml
```diff
@@ -70,6 +70,22 @@ disable_target_static_assertions = [
 ]
 # If this is active all the code compiles in synchronous mode. Otherwise `async` is supported.
 sync-api = ["ac-compose-macros/sync-api", "maybe-async/is_sync"]
+# Use the `jsonrpsee` crate for websocket communication. Does provide sync and async support but needs a tokio runtime.
+# Provides convenience functions such as subscription callbacks.
+jsonrpsee-client = ["std", "jsonrpsee", "futures"]
+# Use the `tungstenite` crate for websocket communication. No async support but has some reconnection capabilities.
+tungstenite-client = ["std", "tungstenite"]
+# Use the `ws` crate for websocket communication. No async support.
+# Establishes a new connection for each request and therefore is limited in terms of performance.
```
haerdib, line 79: Cool! I like this documentation very much. Your example linking below just gave me the idea: Would it make sense to link to the ws_client example as well? (same for tungestenite)

haerdib (approved):
Nice, thanks a lot!

Cargo.toml
Niederb: Makes sense. I added the examples to the documentation.

PR #559, opened 2023-04-28
Remove extra `is` from example println

haerdib:


Niederb (approved):
That **is is** a nice improvement 😉

echevrier (approved):
Yes, really nice

PR #565, opened 2023-05-04
Fix CI to run with latest node

echevrier:
no more --ws-external CMD since Substrate commit [71d749](https://github.com/paritytech/substrate/commit/71d749c74e43c7a840c94fcbbdd2b0172a21d473#diff-9290817490e013034857f5b228f357bb0e93d7e33e028cdeea17d20fa1d21b35L86)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,7 +156,7 @@ jobs:
 
       - name: Run latest node
         run: |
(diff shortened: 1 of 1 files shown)


haerdib (approved):
LGTM! can you change the label to `E0`? This is no change in the functional code, so no need to list in the release.

PR #573, opened 2023-05-11
Add function `get_system_account_next_index`

haerdib:
Fixes #572

Regarding the `// FIXME: Remove once #391 is fixed.` : Issue added in #574

PR #575, opened 2023-05-15
Move `get_system_account_next_index` to `GetAccountInformation`

haerdib:
Only realized after merging #573 , but I believe `get_system_account_next_index` fits better to `GetAccountInformation`. This way, `index` and `AccountId` do not need to be redefined for trait `SystemApi`.


Niederb (approved):
I agree. Makes more sense in GetAccountInformation

echevrier (approved):
I agree

PR #585, opened 2023-06-13
Readd `KeystoreExt` to `LocalKeystore`

### LINKED ISSUE #584: Readd `KeystoreExt`
Somewhere inbetween our custom KeystoreExt got lost: https://github.com/scs/substrate-api-client/commit/686b7ef0aa8da255d3864a3fc703e32193813700#diff-6864a99a9567b596d2cf739abccc32e407278735b51103ab33290937a49ecaaaL73-L83

We should readd it as it is still used by some.

Cherry-pick: https://github.com/scs/substrate-api-client/commit/fa4c8314b3fadf814b53dadc44aa2cf595cd2d5b

haerdib:
Readded `KeystoreExt` but in a separate file this time, so it does not get deleted accidentally.

closes #584


masapr (approved):
I'm ok with moving the KeystoreExt to a new file. But to prevent the KeystoreExt to be deleted again, we should rather add a test, that uses the KeystoreExt.

haerdib:
Good input. I will implement a Test doing just that

client-keystore/src/lib.rs
```diff
@@ -36,6 +35,10 @@ use std::{
 
 use sc_keystore::{Error, Result};
 
+mod keystore_ext;
+pub use keystore_ext::KeystoreExt;
```
masapr, line 39: Do we need this line? The KeystoreExt is not used here, right?

masapr (approved):
I looked into testing without using the filesystem. I ended up, writing a simple test for 'generate()', I think, it's well enough tested now. I'm not sure, if it's possible to properly test the 'public_keys()' with an in-memory store...

client-keystore/src/lib.rs
haerdib: Yes, otherwise the `KeystoreExt` won't be accessible from the outside.
We could do it the following way:
```rust
pub mod keystore_ext;
```
but then one would need to import it with `use substrate_client_keystore::keystore_ext::KeystoreExt;`. I prefer the above TBH

PR #587, opened 2023-06-14
support metadata v15

### LINKED ISSUE #544: Support metadata V15
Not yet stable, but it's in the Substrate code. Subxt has already followed: https://github.com/paritytech/subxt/pull/912/files

The first step would be to add support for metadata v15. Important: Do not break support for metadata v14.

masapr:
Support metadata v15. Any metadata v14 is converted to v15.
closes #544


node-api/src/from_v14_to_v15.rs
```diff
@@ -0,0 +1,88 @@
+use frame_metadata::{v14, v15};
```
haerdib, line 1: Can you add a file header?

haerdib (approved):

PR #594, opened 2023-06-22
Make `get_system_account_next_index` available for `no_std`

### LINKED ISSUE #574: Make `get_system_account_next_index` available in no_std
The problem: `AccountId` does not implement Serialize in no_std. Should be solved once #391 is done.

Copying the type from Substrate and implement Serialize for it would be another solution, but I'm not sure if that's the way to go. Depends on https://github.com/scs/substrate-api-client/issues/553

masapr:
removed std feature flag for `get_system_account_next_index`

closes #574

### DIFF
--- a/src/api/rpc_api/frame_system.rs
+++ b/src/api/rpc_api/frame_system.rs
@@ -30,8 +30,6 @@ pub trait GetAccountInformation {
 	type AccountData;
 
 	/// Retrieves the next account index as available on the node.
-	// FIXME: Remove std feature #574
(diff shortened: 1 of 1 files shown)


haerdib (approved):
Finally 🥳 

This PR does not break the api, does it? To my understanding, it simply makes the function available in no_std as well.

masapr:
Yes it doesn't break it, but it changes it. So, I wasn't sure about the label. I'll change it.

PR #597, opened 2023-06-26
Update Metadata, Error and Events

### LINKED ISSUE #307: Enhance std version of api-client
Some features are currently not implemented at all because they are not std compatible. However, it could be made available in std modus only.

Example: metadata cache

### LINKED ISSUE #564: Improve decoding of `EventDetails` and `DispatchErrors`
Similar to subxt, use [scale-decode](https://github.com/paritytech/scale-decode) and [scale-encode](https://github.com/paritytech/scale-encode/blob/main/scale-encode/src/lib.rs) :
- more flexible to substrate changes( new variant, indexes changes, ...)
- using the visitor seems to allow zero-copy decoding.

These libraries need to be made no-std compatible.

Replace the [decoder ](https://github.com/scs/substrate-api-client/tree/master/node-api/src/decoder).

Rework [EventDetails](https://github.com/scs/substrate-api-client/blob/e1c766cb9312cedd40f024d8e4a1d955abc8ab1c/node-api/src/events.rs#L156) and [Dispatch Errors](https://github.com/scs/substrate-api-client/blob/e1c766cb9312cedd40f024d8e4a1d955abc8ab1c/node-api/src/error.rs#L91) decoding

haerdib:
Updates the Metadata, Error and Event to use the new visitor decoding. The following was updated:

### Dispatch Error
- Renamed `ModuleErrorData` to `RawModuleError`. This seems more appropriate. 
- Updated `decode_from` to use the visitor pattern and less verbose code structure.

### Events
- `EventDetails` are now generic over the Hash because `topics` are now used internally
- `EventDetails::decode_from` was changed to use the visitor pattern.
- Introduced `EventMetadataDetails` struct as helper for retrieving pallet specific event variants.

### Metadata
- Removed `pub` access from underlying variables of the `Metadata` struct. Reasoning: Modifying single attributes does not make sense because all parameters are derived from the underlying `runtime_metadata`. So to change one parameter, the runtime_metadata should be updated and impacting the other, dependent parameters.
- Removed `EventMetadata` and `ErrorMetadata`. Reasoning: Events and errors are part of a specific pallet. So these Metadata structs were basically a redundant definition of the pallet metadata.
- Introduced `apis` and `extrinsic` Metadata to follow Metadata v15 changes.
- Introduced a wrapper struct for `PalletMetadata` and `RuntimeApiMetadata`. This exposes the `PortableRegistry` for `TypeInfo` and the possibility to return an Iterator (which needs a lifetime), without having to define the full `PalletMetadataInner` struct with lifetime parameters.
- `PalletMetadata` now exposes the error and event types with `event_variants` and similar functions.

closes #564 
closes #307 because the std-only fields no longer exist either way (see [metadata-type](https://github.com/scs/substrate-api-client/pull/597/files#diff-c5e2a2a9f77365e024c77a2ac3fcbd8f6a0e041434b8f750fdd2c39566950385L48-L50) changes)

### DIFF
--- a/compose-macros/src/lib.rs
+++ b/compose-macros/src/lib.rs
@@ -93,7 +93,7 @@ macro_rules! compose_extrinsic_with_nonce {
 
             debug!("Composing generic extrinsic for module {:?} and call {:?}", $module, $call);
 
-			let metadata = $api.metadata().clone();
+			let metadata = $api.metadata();
             let call = $crate::compose_call!(metadata, $module, $call $(, ($args)) *);
             if let Some(signer) = $api.signer() {
                 $crate::compose_extrinsic_offline!(
--- a/compose-macros/src/lib.rs
+++ b/compose-macros/src/lib.rs
@@ -37,11 +37,11 @@ mod rpc;
 macro_rules! compose_call {
 ($node_metadata: expr, $pallet: expr, $call_name: expr $(, $args: expr) *) => {
         {
-            let pallet = $node_metadata.pallet($pallet).unwrap().to_owned();
+            let pallet = $node_metadata.pallet_by_name($pallet).unwrap().to_owned();
 
-            let call_index = pallet.call_indexes.get($call_name).unwrap();
+            let call_index = pallet.call_variant_by_name($call_name).unwrap().index;
 
-            ([pallet.index, *call_index as u8] $(, ($args)) *)
+            ([pallet.index(), call_index as u8] $(, ($args)) *)
         }
     };
 }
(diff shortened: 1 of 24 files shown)


echevrier (approved):
Really nice. Thanks

compose-macros/src/lib.rs
```diff
@@ -93,7 +93,8 @@ macro_rules! compose_extrinsic_with_nonce {
 
             debug!("Composing generic extrinsic for module {:?} and call {:?}", $module, $call);
 
-            let call = $crate::compose_call!($api.metadata().clone(), $module, $call $(, ($args)) *);
+			let metadata = $api.metadata().clone();
```
Niederb, line 96: Do we need this clone?

Niederb (approved):
Just a small question regarding a clone, but looks good to me 👍

compose-macros/src/lib.rs
haerdib: Good Input. Doesn't look like it does, removed it :)

Issue #590, opened 2023-06-20
Update to upstream scale-bits, scale-decode and scale-encode


haerdib:
Parity has agreed to make these crates `no_std` compatible. Once the PRs have been merged, we should change the scale-... branches back to upstream (currently pointing to a fork), resp. to the newly published version on crates.io

- [x] scale bits has been released in no_std: https://crates.io/crates/scale-bits: v0.4.0
- [x] scale-decode: https://github.com/paritytech/scale-decode/pull/26 (current crates.io version: 0.7.0)
- [x] scale-encode: https://github.com/paritytech/scale-encode/pull/11, merged, but not yet released on crates.io (current version: 0.3.0)

jsdw:
Just to say, scale-decode 0.8 and scale-encode 0.4 are now released with no_std compat; thanks for your contributions! :)

PR #599, opened 2023-06-29
Update scale-bits and scale-encode

haerdib:
Related to #590


echevrier (approved):
Nice. Congratulations.

PR #601, opened 2023-06-29
Take Bytes parameters as borrow

Niederb:
Take `Bytes` parameters by reference. Makes the code a bit more idiomatic and efficient.
We mostly use it for serialization and this works fine as it is also implemented for the reference.


echevrier (approved):
LGTM

haerdib (approved):
🥳 Yeesss, thanks a lot!

PR #602, opened 2023-06-29
fetch more storage keys than the maximum limit set by substrate

### LINKED ISSUE #588: fetch all pages until end for `get_storage_keys_paged`
instead of asking for a max_count, max count could be assumed to be 1000 (seems to be rpc default) and then just fetch all pages directly

echevrier:
Add **get_all_storage_keys_paged_up_to_count** to fetch required number of keys, independantly of substrate limit:
- call several times the substrate RPC call with the max limited number (1000), if required

The default call, **get_storage_keys_paged** performs according to substrate: if required number of keys> 1000, an error is thrown. The documentation warns the user.

This does not completely fix #588. To fetch all pages, you need to know how many keys are stored. Currently,  `count `is mandatory in [`state_getKeysPaged`](https://github.com/paritytech/substrate/blob/fe2d51377fe1a8fbbb0599523f4fdc1422c82f81/client/rpc/src/state/mod.rs#L374) and substrate doesn't support the feature for obtaining all keys.

### DIFF
--- a/examples/examples/get_storage.rs
+++ b/examples/examples/get_storage.rs
@@ -76,7 +76,7 @@ async fn main() {
 	let storage_key_prefix = api.get_storage_map_key_prefix("System", "Account").unwrap();
 	let max_keys = 3;
 	let storage_keys = api
-		.get_storage_keys_paged(Some(storage_key_prefix), max_keys, None, None)
+		.get_storage_keys_paged_limited(Some(storage_key_prefix), max_keys, None, None)
 		.unwrap();
 	assert_eq!(storage_keys.len() as u32, max_keys);
 	// Get the storage values that belong to the retrieved storage keys.
@@ -91,7 +91,7 @@ async fn main() {
 	let storage_double_map_key_prefix =
 		api.get_storage_double_map_key_prefix("Staking", "ErasStakers", 0).unwrap();
 	let double_map_storage_keys = api
-		.get_storage_keys_paged(Some(storage_double_map_key_prefix), max_keys, None, None)
+		.get_storage_keys_paged_limited(Some(storage_double_map_key_prefix), max_keys, None, None)
 		.unwrap();
 
 	// Get the storage values that belong to the retrieved storage keys.
--- a/examples/examples/get_storage.rs
+++ b/examples/examples/get_storage.rs
@@ -76,7 +76,7 @@ async fn main() {
 	let storage_key_prefix = api.get_storage_map_key_prefix("System", "Account").unwrap();
 	let max_keys = 3;
 	let storage_keys = api
-		.get_storage_keys_paged_limited(Some(storage_key_prefix), max_keys, None, None)
+		.get_storage_keys_paged(Some(storage_key_prefix), max_keys, None, None)
 		.unwrap();
 	assert_eq!(storage_keys.len() as u32, max_keys);
 	// Get the storage values that belong to the retrieved storage keys.
@@ -91,7 +91,7 @@ async fn main() {
 	let storage_double_map_key_prefix =
 		api.get_storage_double_map_key_prefix("Staking", "ErasStakers", 0).unwrap();
 	let double_map_storage_keys = api
-		.get_storage_keys_paged_limited(Some(storage_double_map_key_prefix), max_keys, None, None)
+		.get_storage_keys_paged(Some(storage_double_map_key_prefix), max_keys, None, None)
 		.unwrap();
 
 	// Get the storage values that belong to the retrieved storage keys.
--- a/src/api/rpc_api/state.rs
+++ b/src/api/rpc_api/state.rs
@@ -20,6 +20,7 @@ use ac_node_api::MetadataError;
 use ac_primitives::config::Config;
 use alloc::{borrow::ToOwned, string::String, vec, vec::Vec};
 use codec::{Decode, Encode};
+use core::cmp;
 use log::*;
 use serde::de::DeserializeOwned;
 use sp_storage::{StorageChangeSet, StorageData, StorageKey};
@@ -99,7 +100,7 @@ pub trait GetStorage {
 	/// If `start_key` is passed, return next keys in storage in lexicographic order.
 	///
 	/// `at_block`: the state is queried at this block, set to `None` to get the state from the latest known block.
-	async fn get_storage_keys_paged(
+	async fn get_storage_keys_paged_limited(
 		&self,
 		prefix: Option<StorageKey>,
 		count: u32,
@@ -113,7 +114,7 @@ pub trait GetStorage {
 	/// If `start_key` is passed, return next keys in storage in lexicographic order.
 	///
 	/// `at_block`: the state is queried at this block, set to `None` to get the state from the latest known block.
-	async fn get_all_storage_keys_paged_up_to_count(
+	async fn get_storage_keys_paged(
 		&self,
 		prefix: Option<StorageKey>,
 		count: u32,
@@ -267,7 +268,7 @@ where
 		}
 	}
 
-	async fn get_storage_keys_paged(
+	async fn get_storage_keys_paged_limited(
 		&self,
 		storage_key_prefix: Option<StorageKey>,
 		count: u32,
@@ -284,7 +285,7 @@ where
 		Ok(storage)
 	}
 
-	async fn get_all_storage_keys_paged_up_to_count(
+	async fn get_storage_keys_paged(
 		&self,
 		storage_key_prefix: Option<StorageKey>,
 		count: u32,
@@ -292,22 +293,26 @@ where
 		at_block: Option<Self::Hash>,
 	) -> Result<Vec<StorageKey>> {
 		let mut storage_keys: Vec<StorageKey> = Vec::new();
-		let mut still_todo = count;
-		let mut new_key = start_key;
+		let mut keys_left_to_fetch = count;
+		let mut new_start_key = start_key;
 
-		while still_todo > 0 {
-			let new_count = if still_todo < STORAGE_KEYS_PAGED_MAX_COUNT {
-				still_todo
-			} else {
-				STORAGE_KEYS_PAGED_MAX_COUNT
-			};
-
+		while keys_left_to_fetch > 0 {
+			let new_count = cmp::min(STORAGE_KEYS_PAGED_MAX_COUNT, keys_left_to_fetch);
 			let mut keys = self
-				.get_storage_keys_paged(storage_key_prefix.clone(), new_count, new_key, at_block)
+				.get_storage_keys_paged_limited(
+					storage_key_prefix.clone(),
+					new_count,
+					new_start_key,
+					at_block,
+				)
 				.await?;
+			let num_keys = keys.len() as u32;
 			storage_keys.append(&mut keys);
-			still_todo -= new_count;
-			new_key = keys.last().map(|x| x.to_owned());
+			if num_keys < new_count {
+				break
+			}
+			keys_left_to_fetch -= new_count;
+			new_start_key = keys.last().map(|x| x.to_owned());
 		}
 
 		Ok(storage_keys)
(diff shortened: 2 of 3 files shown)


src/api/rpc_api/state.rs
```diff
@@ -100,6 +107,20 @@ pub trait GetStorage {
 		at_block: Option<Self::Hash>,
 	) -> Result<Vec<StorageKey>>;
 
+	/// Retrieve up to `count` keys. Support prefix and pagination support.
+	/// The number of keys returned is not limited. For big numbers, the rpc calls will be made several times.
+	/// Up to `count` keys will be returned.
+	/// If `start_key` is passed, return next keys in storage in lexicographic order.
+	///
+	/// `at_block`: the state is queried at this block, set to `None` to get the state from the latest known block.
+	async fn get_all_storage_keys_paged_up_to_count(
```
haerdib, line 116: That's how I imagined this function. But I think this version should be the default, not the original one that throws an error. 

So I'm in favour of renaming this function to `get_all_storage_keys_paged` and rename the original to something else. Not sure if we should use `get_all_storage_keys_paged_up_to_count` as a name, because that's true for both, new and original, functions.

src/api/rpc_api/state.rs
```diff
@@ -263,6 +284,35 @@ where
...
 
+	async fn get_all_storage_keys_paged_up_to_count(
+		&self,
+		storage_key_prefix: Option<StorageKey>,
+		count: u32,
+		start_key: Option<StorageKey>,
+		at_block: Option<Self::Hash>,
+	) -> Result<Vec<StorageKey>> {
+		let mut storage_keys: Vec<StorageKey> = Vec::new();
+		let mut still_todo = count;
```
haerdib, line 295:
```suggestion
		let mut keys_left_to_fetch = count;
```

src/api/rpc_api/state.rs
```diff
@@ -263,6 +284,35 @@ where
...
+		let mut new_key = start_key;
+
+		while still_todo > 0 {
```
Niederb, line 298: Can't we get stuck in this loop if there are fewer events available than `count`?

haerdib, line 298:
```suggestion
		while still_todo >= 0 {
```

src/api/rpc_api/state.rs
```diff
@@ -263,6 +284,35 @@ where
...
+	async fn get_all_storage_keys_paged_up_to_count(
+		&self,
+		storage_key_prefix: Option<StorageKey>,
+		count: u32,
+		start_key: Option<StorageKey>,
+		at_block: Option<Self::Hash>,
+	) -> Result<Vec<StorageKey>> {
+		let mut storage_keys: Vec<StorageKey> = Vec::new();
+		let mut still_todo = count;
+		let mut new_key = start_key;
```
haerdib, line 296:
```suggestion
		let mut new_start_key = start_key;
```

testing/examples/state_tests.rs
```diff
@@ -89,13 +89,23 @@ async fn main() {
 	let _keys = api.get_keys(storage_key, None).unwrap().unwrap();
 	let _constants: Balance = api.get_constant("Balances", "ExistentialDeposit").unwrap();
 
-	let max_keys = 3;
+	let max_keys = 2003;
+	let result =
+		api.get_storage_keys_paged(Some(storage_key_prefix.clone()), max_keys.clone(), None, None);
+	assert!(result.is_err());
```
haerdib, line 95: I don't think we need to check for an error here. That's an error from Substrate itself, which may change over time

testing/examples/state_tests.rs
```diff
@@ -89,13 +89,23 @@ async fn main() {
...
+	assert!(format!("{result:?}").contains("count exceeds maximum value"));
+
 	let storage_keys = api
-		.get_storage_keys_paged(Some(storage_key_prefix), max_keys, None, None)
+		.get_all_storage_keys_paged_up_to_count(Some(storage_key_prefix), max_keys, None, None)
 		.unwrap();
-	assert_eq!(storage_keys.len() as u32, max_keys);
+	assert!(storage_keys.len() as u32 > 3);
+	assert!(storage_keys.len() as u32 <= max_keys);
```
haerdib, line 102: Any chance to assert that it is equal to `max_keys`? Only then we will know that our function is working correctly. 
Maybe change the prefix to `None` ?

Niederb (commented):
I'm not quite sure if it makes sense to have `get_all_storage_keys_paged_up_to_count`  as a separate function. It can do everything that `get_storage_keys_paged` can, so for me it should become the new `get_storage_keys_paged` and `get_storage_keys_paged` could become an internal helper method.

src/api/rpc_api/state.rs
```diff
@@ -263,6 +284,35 @@ where
...
+
+		while still_todo > 0 {
+			let new_count = if still_todo < STORAGE_KEYS_PAGED_MAX_COUNT {
+				still_todo
+			} else {
+				STORAGE_KEYS_PAGED_MAX_COUNT
+			};
+
+			let mut keys = self
```
haerdib, line 305: We need somewhere a "breakout" check. `get_storage_keys_paged` either returns up to the given `max_count` or all the keys with the given prefix, if there's less and `max_count`. So if there are fewer keys returned than the given count, we can break the loop - no more keys left to fetch.

src/api/rpc_api/state.rs
echevrier: I think the default behaviour should be similar to that of the substrate, i.e. the number of keys returned is limited and return an error on big value of count. Users can choose to use the alternative method if they wish to bypass the limits set by substrate. 
get_all_storage_keys_paged is only returning up to count keys, if the count input value < limit. But, yes the name could be more descriptive. How about `get_all_storage_keys_paged_without_limit` or `get_all_storage_keys_paged_workaround`?

haerdib: It's just that I, while using this function, promptly run into this error quite unexpectedly and it took me some time to figure out why. I find it counterintuitive to allow a count input (as `u32` nonetheless) and limit it to `1000` during runtime (which is not documented at all: https://polkadot.js.org/docs/kusama/rpc#getkeyspagedkey-storagekey-count-u32-startkey-storagekey-at-blockhash-vecstoragekey)

So I'm not sure if we should follow the Substrate way here - I believe we can do better. Or what would be the downside of having the "workaround" as a default? Sure, it may take a long time if the user wants to fetch more than 1000 keys. In the worst case, the program even needs to be terminated. But in case an error is returned (due to count > 1000) that probably also ends in a non-functional program that needs to be terminated.

I'm not set on my idea, so if you think it's better to leave the original as default, that's fine to me.

echevrier: > So I'm not sure if we should follow the Substrate way here - I believe we can do better. Or what would be the downside of having the "workaround" as a default? 
> 
That the user is unaware of using a workaround? 

>Sure, it may take a long time if the user wants to fetch more than 1000 keys. In the worst case, the program even needs to be terminated.

If this is the biggest problem our function can cause, then yes, our workaround can be the default function.

haerdib: @Niederb  What's your opinion here?

testing/examples/state_tests.rs
haerdib: That's true, good input.

testing/examples/state_tests.rs
haerdib: Good question.. I don't know 😨  I'd simply put a random number above 1000 and see what the assert statement prints

src/api/rpc_api/state.rs
```diff
@@ -263,6 +284,35 @@ where
...
+		count: u32,
+		start_key: Option<StorageKey>,
+		at_block: Option<Self::Hash>,
+	) -> Result<Vec<StorageKey>> {
+		let mut storage_keys: Vec<StorageKey> = Vec::new();
+		let mut still_todo = count;
+		let mut new_key = start_key;
+
+		while still_todo > 0 {
+			let new_count = if still_todo < STORAGE_KEYS_PAGED_MAX_COUNT {
```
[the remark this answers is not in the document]
Niederb: Yes, there is [`core::cmp::Ord`](https://doc.rust-lang.org/stable/core/cmp/trait.Ord.html)

src/api/rpc_api/state.rs
Niederb: I would also make the new implementation the default. Also see my separate review comment

echevrier:
> I'm not quite sure if it makes sense to have `get_all_storage_keys_paged_up_to_count` as a separate function. It can do everything that `get_storage_keys_paged` can, so for me it should become the new `get_storage_keys_paged` and `get_storage_keys_paged` could become an internal helper method.

We provide both methods. See: https://github.com/scs/substrate-api-client/issues/588#issuecomment-1614619519

src/api/rpc_api/state.rs
echevrier: Ok, then:
`get_storage_keys_paged ` is our workaround and  `get_storage_keys_paged_limited` is the original one, calling the RPC substrate storage_keys_paged only once.

Niederb (approved):
LGTM

src/api/rpc_api/state.rs
```diff
@@ -88,6 +93,23 @@ pub trait GetStorage {
 	) -> Result<Option<V>>;
 
 	/// Retrieve the keys with prefix with pagination support.
+	/// Call the RPC substrate storage_keys_paged, which limits the number of returned keys.
+	/// See https://github.com/paritytech/substrate/blob/9f6fecfeea15345c983629af275b1f1702a50004/client/rpc/src/state/mod.rs#L54
```
haerdib, line 97: Can you move this description `See https://github.com/paritytech/substrate/blob/9f6fecfeea15345c983629af275b1f1702a50004/client/rpc/src/state/mod.rs#L54`

to the bottom with only two backslashes? I don't think the link should show up in the function description

testing/examples/state_tests.rs
```diff
@@ -89,13 +89,26 @@ async fn main() {
...
+		None,
+	);
+	assert!(result.is_err());
+	assert!(format!("{result:?}").contains("count exceeds maximum value"));
+
 	let storage_keys = api
 		.get_storage_keys_paged(Some(storage_key_prefix), max_keys, None, None)
 		.unwrap();
-	assert_eq!(storage_keys.len() as u32, max_keys);
+	assert_eq!(storage_keys.len() as u32, 13);
```
haerdib, line 105: Hmm.. that's too bad. So we basically don't fully test our function loop currently. So I suppose we should introduce a  unit test. But for this we need a mock. Could be done in a separate issue. What do you think?

examples/examples/get_storage.rs
```diff
@@ -91,7 +91,7 @@ async fn main() {
 	let storage_double_map_key_prefix =
 		api.get_storage_double_map_key_prefix("Staking", "ErasStakers", 0).unwrap();
 	let double_map_storage_keys = api
-		.get_storage_keys_paged(Some(storage_double_map_key_prefix), max_keys, None, None)
+		.get_storage_keys_paged_limited(Some(storage_double_map_key_prefix), max_keys, None, None)
```
haerdib, line 94: Why not leave this as is? (same for be function above)

testing/examples/state_tests.rs
echevrier: Add issue: https://github.com/scs/substrate-api-client/issues/605

src/api/rpc_api/state.rs
```diff
@@ -88,6 +93,23 @@ pub trait GetStorage {
 	) -> Result<Option<V>>;
 
 	/// Retrieve the keys with prefix with pagination support.
+	/// Call the RPC substrate storage_keys_paged, which limits the number of returned keys.
+	///
+	/// Up to `count` keys will be returned. If `count` is too big, an error will be returned
+	/// If `start_key` is passed, return next keys in storage in lexicographic order.
+	///
+	/// `at_block`: the state is queried at this block, set to `None` to get the state from the latest known block.
+	// See https://github.com/paritytech/substrate/blob/9f6fecfeea15345c983629af275b1f1702a50004/client/rpc/src/state/mod.rs#L54
```
haerdib, line 102: ❤️

haerdib (approved):
Awesome, thanks a lot!

Issue #506, opened 2023-03-31
Investigate: Is the `TransferEventArgs` in the example obsolete?


haerdib:
Question: is the definition here:
https://github.com/scs/substrate-api-client/blob/10084fa48c77a16d78e36c48c1af2404c141b85b/examples/examples/event_error_details.rs#L28-L38

obsolete? If so, we should remove it.

PR #603, opened 2023-06-30
Remove obsolete TransferEventArgs

echevrier:
TransferEventArgs was needed when we checked the transition with events (`wait_for_event`). No longer used
Closes https://github.com/scs/substrate-api-client/issues/506


Niederb (approved):
👍

Issue #586, opened 2023-06-14
Investigate: Rename `SubstrateKitchensinkConfig` to something more generic


haerdib:
From https://github.com/encointer/encointer-node/pull/328#discussion_r1229303759:

>But it seems to be generic enough such that it works for us too. So it could be something like `DefaultAssetRuntimeConfig`.

I agree with that line of thought. SubstratKitchensinkConfig is correct. But maybe a little too confining.

echevrier:
According to Substrate, the node example runtime is NOTHING production-like, and should not be used, nor inspired from ;) That's why they've renamed the node-runtime to node-kitchensink-runtime. 
Changing the name of our configuration is then deceptively reassuring,  but I can understand, that for live nodes, referring to a Kitchensink is not so nice.

We have another config  that is compatible with a polkadot node: [PolkadotConfig  ](https://github.com/scs/substrate-api-client/blob/master/primitives/src/config/polkadot.rs#L16C18-L16C18). If we rename SubstratKitchensinkConfig without reference to Substrate, we should also rename PolkadotConfig.

A little too confining? I'm not sure where to put the emphasis. For the moment the differences between the configs are the methods of payment or the nodes that use them. If we don't want to make any reference to node-kitchensink-runtime, I'd like to remain as neutral as possible. So maybe DefaultAssetRuntimeConfig instead of SubstratKitchensinkConfig and DefaultRuntimeConfig instead of PolkadotConfig?

Other possibilities 
- DefaultSubstrateAssetRuntimeConfig and DefaultPolkadotRuntimeConfig? 
- SubstrateAssetRuntimeConfig and PoladotRuntimeConfig?

@haerdib , @clangenb What do you think?

clangenb:
Yes, I can understand the rationale why they renamed it into the kitchensink runtime. However, if the config works for polkadot and kusama, I think it shouldn't have the kitchensink name in it. I can agree with your second set of suggestions, but would like to add some more variants:

* `AssetRuntimeConfig`, `PolkadotConfig`

I don't have a strong tendency towards your or mine suggestions.

haerdib:
Maybe I'm a little behind here, but for me, the difference between `Polkadot` and `Substrate` is not clear in regard to the config. Both, Substrate and Polkadot nodes may have both types of Payments, and both, substrate-node-template (not kitchsink-runtime, as @echevrier noted should not be used as template), and polkadot nodes by default use plain tip. 

[...]

PR #604, opened 2023-07-04
Rename runtime configs to something more generic

echevrier:
Rename SubstratKitchensinkConfig into AssetRuntimeConfig 
Rename PolkadotConfig into DefaultRuntimeConfig
Close #https://github.com/scs/substrate-api-client/issues/586


primitives/src/config/asset_runtime_config.rs
```diff
@@ -17,9 +17,9 @@ use core::fmt::Debug;
 
 /// Default set of commonly used types by Substrate kitchensink runtime.
 #[derive(Decode, Encode, Clone, Eq, PartialEq, Debug)]
-pub struct SubstrateKitchensinkConfig {}
```
haerdib, line 20: Cool, thanks a lot! Too avoid a lot of confusion: What would you think about defining a "new" type that has the same name as the old one, but marked with a deprecated?

E.g. 
```rust
#[deprecated(note="Please use `AssetRuntimeConfig` instead, this will be removed soon.")]
pub type SubstrateKitchensinkConfig  = AssetRuntimeConfig;
```

Same for the `PolkadotConfig`.

haerdib: Not sure if the "since" works for future releases, but it would certainly help us to note how old the deprecated text is and when to remove it.

Niederb (approved):
😄

haerdib (approved):
🥳

Issue #592, opened 2023-06-21
Is `ExtrinsicReport` as Return value for `until` and `until_success` clear enough?


haerdib:
Currently, both functions `watch_until` and `watch_until_sucess` are returning an `ExtrinsicReport`. However, only `until_sucess` returns Events. That only pointed out with comments.
It would be clearer / fail safer if there'd be two Structs.. one returned frum `until` (w.o. Events`) and one with Events, probably a Wrapper struct around the original `ExtrinsicReport`.

See question: https://github.com/scs/substrate-api-client/issues/591#issuecomment-1600543312

echevrier:
Isn't the question more: Is the difference between **`submit_and_watch_extrinsic_until_success`** and **`submit_and_watch_extrinsic_until`** clear enough?

What does **until_success** mean? What do we want to see in the report?

Currently both methods generate an ExtrinsicReport or throw an error when a certain status is reached.  `submit_and_watch_extrinsic_until` allows all status, `submit_and_watch_extrinsic_until_success` only **InBlock** or **Finalized**. Furthermore, only `submit_and_watch_extrinsic_until_success` add the events in the report.
So:
- does success mean an extrinsic is InBlock or Finalized?
- does success mean, we can check the events because the extrinsic succesfully reached the expected status?

Why does `submit_and_watch_extrinsic_until` not fetch the events? If no events were triggered by the extrinsic, then no event are in the report. But if an event has been triggered, then it should be in the report. There is no need to modify the extrinsic report.

haerdib:
So you're suggesting to merge the two functions? E.g. when either InBlock or Finalized is reached, then the Events should be returned? I actually like that idea.. it's way more failsafe. If someone waits for one of these two statuses, one would also like to know if the xt has been successful.

To answer your questions:
- success means that the extrinisc has been successful. A failed extrinsic (e.g. transfer failed because of too little money), may still be "Finalized". But in its failed state.
- Getter calls take extra time and Events take time. In one block there might be a lot of Events, which need to be filtered through to actually get the associated events. So `submit_and_watch_extrinsic_until` is basically the "lightweight" function

echevrier:
So we cannot wait until success. Just hope ;)  
I will let two functions but one report: `submit_and_watch_extrinsic_until `and `submit_and_watch_extrinsic_until_light`and document the difference, i.e. events are not reported for performance reasons. Both functions will accept all status 
[...]

Issue #591, opened 2023-06-21
improve examples to check for extrinsic-success and report errors and events triggered by the extrinsic


brenzi:
https://github.com/scs/substrate-api-client/blob/b884bf3de5c4e984b9c806ccdc96b9fe77cfdb91/examples/examples/transfer_with_ws_client.rs#L70
currently, the example just returns the tx hash. What I'd like to see is:
* did the extrinsic not only get included, but did it yield a extrinsicSuccess event?
* if not, report error
* if events were triggered by this extrinsic, return events

haerdib:
`submit_and_watch_extrinsic_until` currently returns an `ExtrinsicReport`:

https://github.com/scs/substrate-api-client/blob/b884bf3de5c4e984b9c806ccdc96b9fe77cfdb91/src/api/mod.rs#L35-L50

So you can retrieve the `tx_hash`, as well as the last known status or Block hash (if watched long enough).

However, what you want is probably the following function:
https://github.com/scs/substrate-api-client/blob/b884bf3de5c4e984b9c806ccdc96b9fe77cfdb91/src/api/rpc_api/author.rs#L146-L165

Check out the example:
https://github.com/scs/substrate-api-client/blob/b884bf3de5c4e984b9c806ccdc96b9fe77cfdb91/examples/examples/event_error_details.rs#L65-L80

haerdib:
But I agree: the example should be renamed - event_error_details is not appropriate anymore.

haerdib:
The `transfer` examples are mainly to show the different ws clients and I'd like to avoid too many duplicate examples, so printing and checking everything there seems to be an overkill.

Proposing the following:
- change transfer examples ([transfer_with_ws_client.rs](https://github.com/scs/substrate-api-client) /blob/b884bf3de5c4e984b9c806ccdc96b9fe77cfdb91/examples/examples/transfer_with_ws_client.rs) and [transfer_with_ws_client.rs](https://github.com/scs/substrate-api-client/blob/b884bf3de5c4e984b9c806ccdc96b9fe77cfdb91/examples/examples/transfer_with_tungstenite_client.rs)
to use `until_success`. With `unwrap()` to ensure it has been successful, as this is probably the default use case.
- Rename `event_error_details.rs` to `check_for_failed_extrinisc.rs` (to be discussed, I suppose) and update the  `println!` msg.
[...]

PR #606, opened 2023-07-04
Rework submit and watch extrinsics

echevrier:
Rename functions to clarify that events are not always listed in the extrinisc report:
- Light version: `submit_and_watch_extrinsic_until_without_events`: submit and watch the extrinsic until the desired status is reached. No events are fetched, so no events are listed in the report. 
- `submit_and_watch_extrinsic_until`: submit and watch the extrinsic until the desired status is reached. When the extrinisc is InBlock or Finalized, the events are retrieved and added in the report.

**Caution**: The previous function `submit_and_watch_extrinsic_until_success ` is deprecated, but `submit_and_watch_extrinsic_until` has different behaviour. Previously it didn't fetch the events. Now it does so for InBlock and Finalized status

Closes https://github.com/scs/substrate-api-client/issues/592

### DIFF
--- a/examples/examples/compose_extrinsic_offline.rs
+++ b/examples/examples/compose_extrinsic_offline.rs
@@ -66,7 +66,7 @@ async fn main() {
 
 	// Send and watch extrinsic until in block (online).
 	let block_hash = api
-		.submit_and_watch_extrinsic_until(xt, XtStatus::InBlock)
+		.submit_and_watch_extrinsic_until_without_events(xt, XtStatus::InBlock)
 		.unwrap()
 		.block_hash
 		.unwrap();
--- a/examples/examples/contract_instantiate_with_code.rs
+++ b/examples/examples/contract_instantiate_with_code.rs
@@ -21,7 +21,7 @@ use sp_keyring::AccountKeyring;
 use substrate_api_client::{
 	ac_compose_macros::primitives::AssetRuntimeConfig, ac_node_api::StaticEvent,
 	ac_primitives::ExtrinsicSigner, extrinsic::ContractsExtrinsics, rpc::JsonrpseeClient, Api,
-	SubmitAndWatch, SubmitAndWatchUntilSuccess, XtStatus,
+	SubmitAndWatch, XtStatus,
 };
 
 // To test this example in CI, we run it against the Substrate kitchensink node. Therefore, we use the AssetRuntimeConfig
--- a/examples/examples/contract_instantiate_with_code.rs
+++ b/examples/examples/contract_instantiate_with_code.rs
@@ -69,7 +69,7 @@ async fn main() {
 	);
 
 	println!("[+] Creating a contract instance with extrinsic:\n\n{:?}\n", xt);
-	let report = api.submit_and_watch_extrinsic_until_success(xt, false).unwrap();
+	let report = api.submit_and_watch_extrinsic_until(xt, XtStatus::InBlock).unwrap();
 	println!("[+] Extrinsic is in Block. Hash: {:?}\n", report.block_hash.unwrap());
 
 	println!("[+] Waiting for the contracts.Instantiated event");
@@ -89,6 +89,8 @@ async fn main() {
 	let xt = api.contract_call(contract.into(), 500_000, 500_000, vec![0u8]);
 
 	println!("[+] Calling the contract with extrinsic Extrinsic:\n{:?}\n\n", xt);
-	let report = api.submit_and_watch_extrinsic_until(xt, XtStatus::Finalized).unwrap();
+	let report = api
+		.submit_and_watch_extrinsic_until_without_events(xt, XtStatus::Finalized)
+		.unwrap();
 	println!("[+] Extrinsic got finalized. Extrinsic Hash: {:?}", report.extrinsic_hash);
 }
--- a/src/api/rpc_api/author.rs
+++ b/src/api/rpc_api/author.rs
@@ -106,13 +106,21 @@ pub trait SubmitAndWatch {
 
 	/// Submit an extrinsic and watch it until the desired status
 	/// is reached, if no error is encountered previously.
+	/// The events are not fetched. So no events are listed in the report.
+	/// To fetch the triggered events, please use submit_and_watch_extrinsic_until.
 	/// Upon success, a report containing the following information is returned:
 	/// - extrinsic hash
 	/// - if watched until at least `InBlock`:
 	///   hash of the block the extrinsic was included in
 	/// - last known extrinsic (transaction) status
+	/// - no events
 	/// This method is blocking.
-	async fn submit_and_watch_extrinsic_until<Address, Call, Signature, SignedExtra>(
+	async fn submit_and_watch_extrinsic_until_without_events<
+		Address,
+		Call,
+		Signature,
+		SignedExtra,
+	>(
 		&self,
 		extrinsic: UncheckedExtrinsicV4<Address, Call, Signature, SignedExtra>,
 		watch_until: XtStatus,
@@ -125,13 +133,15 @@ pub trait SubmitAndWatch {
 
 	/// Submit an encoded, opaque extrinsic and watch it until the desired status
 	/// is reached, if no error is encountered previously.
+	/// The events are not fetched. So no events are listed in the report.
+	/// To fetch the triggered events, please use submit_and_watch_opaque_extrinsic_until_success.
 	/// Upon success, a report containing the following information is returned:
 	/// - extrinsic hash
 	/// - if watched until at least `InBlock`:
 	///   hash of the block the extrinsic was included in
 	/// - last known extrinsic (transaction) status
 	/// This method is blocking.
-	async fn submit_and_watch_opaque_extrinsic_until(
+	async fn submit_and_watch_opaque_extrinsic_until_without_events(
 		&self,
 		encoded_extrinsic: &Bytes,
 		watch_until: XtStatus,
@@ -153,6 +163,7 @@ pub trait SubmitAndWatchUntilSuccess {
 	/// - last known extrinsic (transaction) status
 	/// - associated events of the extrinsic
 	/// This method is blocking.
+	#[deprecated(note = "please use `submit_and_watch_extrinsic_until` instead")]
 	async fn submit_and_watch_extrinsic_until_success<Address, Call, Signature, SignedExtra>(
 		&self,
 		extrinsic: UncheckedExtrinsicV4<Address, Call, Signature, SignedExtra>,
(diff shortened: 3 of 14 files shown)


examples/examples/compose_extrinsic_offline.rs
```diff
@@ -66,7 +66,7 @@ async fn main() {
 
 	// Send and watch extrinsic until in block (online).
 	let block_hash = api
-		.submit_and_watch_extrinsic_until(xt, XtStatus::InBlock)
+		.submit_and_watch_extrinsic_until_without_events(xt, XtStatus::InBlock)
```
haerdib, line 69: I'd use the new `submit_and_watch_extrinsic_until` here.

examples/examples/contract_instantiate_with_code.rs
```diff
@@ -89,6 +89,8 @@ async fn main() {
 	let xt = api.contract_call(contract.into(), 500_000, 500_000, vec![0u8]);
 
 	println!("[+] Calling the contract with extrinsic Extrinsic:\n{:?}\n\n", xt);
-	let report = api.submit_and_watch_extrinsic_until(xt, XtStatus::Finalized).unwrap();
+	let report = api
```
haerdib, line 92: Leave as is: Normal users want to know if xt was successful or not. Same goes for all the examples I believe.

src/api/rpc_api/author.rs
```diff
@@ -106,13 +106,21 @@ pub trait SubmitAndWatch {
 
 	/// Submit an extrinsic and watch it until the desired status
 	/// is reached, if no error is encountered previously.
+	/// The events are not fetched. So no events are listed in the report.
```
haerdib, line 109:
```suggestion
	/// Associated events are not fetched.
```

src/api/rpc_api/author.rs
```diff
@@ -106,13 +106,21 @@ pub trait SubmitAndWatch {
...
+	/// To fetch the triggered events, please use submit_and_watch_extrinsic_until.
 	/// Upon success, a report containing the following information is returned:
 	/// - extrinsic hash
 	/// - if watched until at least `InBlock`:
 	///   hash of the block the extrinsic was included in
 	/// - last known extrinsic (transaction) status
+	/// - no events
```
haerdib, line 116: Does it make sense to list `no events` in a list where it says ` the following information is returned:` ?

src/api/rpc_api/author.rs
```diff
@@ -106,13 +106,21 @@ pub trait SubmitAndWatch {
...
 	/// This method is blocking.
-	async fn submit_and_watch_extrinsic_until<Address, Call, Signature, SignedExtra>(
+	async fn submit_and_watch_extrinsic_until_without_events<
```
haerdib, line 118: Can you move this function to the bottom of the definitions? I think `submit_and_watch_extrinsic_until` should be the most visible one.

src/api/rpc_api/author.rs
```diff
@@ -174,11 +188,52 @@ pub trait SubmitAndWatchUntilSuccess {
...
+	/// is reached, if no error is encountered previously.
+	/// When status is InBlock or Finalized, the triggered events are listed in the report.
+	/// Returns and error if the extrinsic was not successfully executed.
+	/// If it was successful, a report containing the following is returned:
+	/// - extrinsic hash
+	/// - hash of the block the extrinsic was included in
+	/// - last known extrinsic (transaction) status
+	/// - associated events of the extrinsic (only for InBlock or Finalized)
+	/// This method is blocking.
+	async fn submit_and_watch_extrinsic_until<Address, Call, Signature, SignedExtra>(
```
haerdib, line 211: Move this to top?

src/api/rpc_api/author.rs
```diff
@@ -174,11 +188,52 @@ pub trait SubmitAndWatchUntilSuccess {
...
+	)]
 	async fn submit_and_watch_opaque_extrinsic_until_success(
 		&self,
 		encoded_extrinsic: &Bytes,
 		wait_for_finalized: bool,
 	) -> Result<ExtrinsicReport<Self::Hash>>;
+
+	/// Submit an extrinsic and watch it until the desired status
+	/// is reached, if no error is encountered previously.
+	/// When status is InBlock or Finalized, the triggered events are listed in the report.
```
haerdib, line 203:
```suggestion
```
That information is mentioned in the list below already

src/api/rpc_api/author.rs
```diff
@@ -174,11 +188,52 @@ pub trait SubmitAndWatchUntilSuccess {
...
+	/// Returns and error if the extrinsic was not successfully executed.
```
haerdib, line 204: Suggestion: Split description in two parts:

Something like this:
```rust
/// Submit an extrinsic and watch it until the desired status
/// is reached, if no error is encountered previously.
/// If watched until `InBlock` or `Finalized` this function will either return an Error
/// if the extrinsic execution has failed onchain. 
/// If it was successful, a report containing the following is returned:
/// - extrinsic hash
/// - hash of the block the extrinsic was included in
/// - last known extrinsic (transaction) status
/// - associated events of the extrinsic
/// If not watched until atleast `InBlock` this function will not know if the extrinsic
/// was executed onchain or not and will therefore not return an error upon failed execution.
/// An Error will be returned if the extrinsic has failed to be sent or was not 
/// included into the transaction pool of the node.
/// If no error is enocuntered, a report containing the following is returned:
/// - extrinsic hash
/// - last known extrinsic (transaction) status

```

src/api/rpc_api/author.rs
```diff
@@ -284,11 +347,13 @@ where
 		Signature: Encode,
 		SignedExtra: Encode,
 	{
-		self.submit_and_watch_opaque_extrinsic_until_success(
-			&extrinsic.encode().into(),
-			wait_for_finalized,
-		)
-		.await
+		let xt_status = match wait_for_finalized {
```
haerdib, line 350: no deprecated status needed in the function impl?

testing/examples/author_tests.rs
```diff
@@ -67,50 +67,52 @@ async fn main() {
 
 	thread::sleep(Duration::from_secs(6)); // Wait a little to avoid transaction too low priority error.
 	let xt2 = api.balance_transfer_allow_death(bob.clone(), 1000);
-	let report = api.submit_and_watch_extrinsic_until(xt2, XtStatus::Ready).unwrap();
+	let report = api
+		.submit_and_watch_extrinsic_until_without_events(xt2, XtStatus::Ready)
+		.unwrap();
 	assert!(report.block_hash.is_none());
-	println!("Success: submit_and_watch_extrinsic_until Ready");
+	println!("Success: submit_and_watch_extrinsic_until_without_events Ready");
```
haerdib, line 74: I'd leave the function testing as is. `submit_and_watch_extrinsic_until_without_events ` is automatically tested internally when calling `submit_and_watch_extrinsic_until`. So I'd rather test the wrapper function instead of the internal one.

haerdib: Though, implement one or two calls, just as a showcase for us and others on how to use the function.

src/api/rpc_api/author.rs
echevrier: In the report we will have `events: None`.  
This should make clear that `events: None` in the report, doesn't mean that no events were triggered, but that the events were not fetched. Perhaps it is not clear enough.  Maybe  "- events: None, because events are not fetched " ?

examples/examples/compose_extrinsic_offline.rs
echevrier: Why? We don't check the report/events

examples/examples/contract_instantiate_with_code.rs
echevrier: `api.submit_and_watch_extrinsic_until` is now `api.submit_and_watch_extrinsic_until_without_events`.
`api.submit_and_watch_extrinsic_until` now fetch the events

src/api/rpc_api/author.rs
echevrier: `submit_and_watch_extrinsic_until` belongs to a different trait.

examples/examples/contract_instantiate_with_code.rs
haerdib: Yes, I know. But this is an example for the users - so we should use what we believe is best practice. And personally, I'd recommend to use `submit_and_watch_extrinsic_until`.

src/api/rpc_api/author.rs
echevrier: I think it is important to note that no events are listed in the report.

src/api/rpc_api/author.rs
haerdib: How the report looks like can be looked up easily in the code. So I don't think it's necessary to mention twice (already written above), that event are not fetched

src/api/rpc_api/author.rs
echevrier: I don't think so.

src/api/rpc_api/author.rs
haerdib: Oh Right. Since we're removing the `success` function, we should also rethink our Trait structure. I'd vote for moving the new function to `SubmitAndWatch` and remove `SubmitAndWatchUntilSuccess` completely once the deprecated functions are removed. I don't see the use case in having two different traits.

examples/examples/contract_instantiate_with_code.rs
echevrier: I don't think we should hide `submit_and_watch_extrinsic_until_without_events`. Why should a user call `submit_and_watch_extrinsic_until` if he is not interested in events?
In any case, we should rethink this in another problem: #591

haerdib: It's not just about events, but mostly about the Error part - `submit_and_watch_extrinsic_until_without_events` will not error, even if the extrinsic has failed onchain. This is fairly misleading for the unaware user.

But we can do that in #591, agreed.

src/api/rpc_api/author.rs
haerdib: I'd move the new `submit_and_watch_extrinsic_until` already now to the `SubmitAndWatch` trait. Otherwise, upon trait removal, users will have to change the trait import to keep using the function.  Or is there a downside to already implement the function in that trait?

testing/examples/author_tests.rs
echevrier: ?
We have in this test:
let events = api.submit_and_watch_extrinsic_until(xt6, XtStatus::InBlock)

haerdib: The idea of this crate was to test every possible outcome of each function in the api-client, that was not unit testable (due to Substrate node dependency). 

For example, one case that is currently not tested, is if ` api.submit_and_watch_extrinsic_until` really works for statuses below `InBlock` E.g. does this part of the code
```rust
if watch_until < XtStatus::InBlock {
	return Ok(report)
}
```
really work? Is the report returned without events? Or does it error upon not finding any associated events onchain?

src/api/rpc_api/author.rs
echevrier: Nice to add the errors differences

src/api/rpc_api/author.rs
echevrier: Let's try it ;)

examples/examples/compose_extrinsic_offline.rs
haerdib: See https://github.com/scs/substrate-api-client/pull/606/files/6243e73241a007764a3f889237ee1a224f9e0cd3#r1253142071

src/api/rpc_api/author.rs
```diff
@@ -106,13 +106,77 @@ pub trait SubmitAndWatch {
...
 	/// Submit an extrinsic and watch it until the desired status
 	/// is reached, if no error is encountered previously.
+	///
+	/// If watched until `InBlock` or `Finalized`, this function will
+	/// return an error if the extrinsic was not successfully executed.
+	/// If it was successful, a report containing the following is returned:
+	/// - extrinsic hash
+	/// - hash of the block the extrinsic was included in
+	/// - last known extrinsic (transaction) status
+	/// - associated events of the extrinsic (only for InBlock or Finalized)
```
haerdib, line 116:
```suggestion
	/// - associated events of the extrinsic
```

Optional, since it's already mentioned above. If you'd rather keep it, you should also put it behind the bullet point `hash of the block the extrinsic was included in` for consistency. I'm fine with either.

src/api/rpc_api/author.rs
```diff
@@ -106,13 +106,77 @@ pub trait SubmitAndWatch {
...
+	///
+	/// If not watched until at least `InBlock`, this function will not know if the extrinsic
+	/// has been executed on chain or not and will therefore not return an error if execution fails..
```
haerdib, line 119:
```suggestion
	/// has been executed on chain or not and will therefore not return an error if execution fails.
```

src/api/rpc_api/author.rs
```diff
@@ -106,13 +106,77 @@ pub trait SubmitAndWatch {
...
+	/// An error will be returned, if the extrinsic has failed to be sent or if it has not been
```
haerdib, line 120:
```suggestion
	/// An error will be returned if the extrinsic has failed to be sent or if it has not been
```

src/api/rpc_api/author.rs
```diff
@@ -125,13 +189,15 @@ pub trait SubmitAndWatch {
 
 	/// Submit an encoded, opaque extrinsic and watch it until the desired status
 	/// is reached, if no error is encountered previously.
+	/// The events are not fetched. So no events are listed in the report.
+	/// To fetch the triggered events, please use submit_and_watch_opaque_extrinsic_until_success.
```
haerdib, line 193:
```suggestion
	/// To fetch the triggered events, please use submit_and_watch_opaque_extrinsic_until.
```

haerdib (approved):
Small comments only, thanks for your patience!

haerdib (approved):
🥳

Niederb (approved):
LGTM

PR #611, opened 2023-07-10
Fix comments of default Configs

haerdib:


primitives/src/config/asset_runtime_config.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
 use codec::{Decode, Encode};
 use core::fmt::Debug;
 
-/// Default set of commonly used types by Substrate kitchensink runtime.
+/// Commonly used config of Substrate runtimes that use the asset pallet.
```
echevrier, line 18: Maybe "Standard runtime config for Substrate and Polkadot nodes that use the asset pallet" ?

primitives/src/config/default_runtime_config.rs
```diff
@@ -11,7 +11,7 @@ use crate::{
 	config::WithExtrinsicParams, AssetRuntimeConfig, Config, GenericExtrinsicParams, PlainTip,
 };
 
-/// Default set of commonly used types by Polkadot nodes.
+/// Commonly used config of Substrate and Polkadot nodes.
 pub type DefaultRuntimeConfig =
```
echevrier, line 15: Similar to previous comment:  "Standard runtime config for Substrate and Polkadot nodes that use native payment" ?

echevrier (changes requested):

primitives/src/config/default_runtime_config.rs
haerdib: I'd refrain from using native payment, as Parity does not use it either. 

What about a simple `/// Standard runtime config for Substrate and Polkadot nodes.` ?

PR #615, opened 2023-07-11
Add EventReport and failed xt check to examples

### LINKED ISSUE #591: improve examples to check for extrinsic-success and report errors and events triggered by the extrinsic

haerdib:
The following has been changed:

## Examples
- All examples:
    -  Updated file description to use `DefaultRuntimeConfig` instead of Polkadot Config and added a little more text.
    - Where used: changed function `submit_and_watch_extrinsic_until_without_event` to `submit_and_watch_extrinsic_until`. Reasoning: `submit_and_watch_extrinsic_until_without_events` does not check if the extrinsic has failed, which might lead to confusion. So I'd suggest to use `submit_and_watch_extrinsic_until` instead, which checks for failure if possible, and if not (due to watching not long enough), it acts the very same as  `submit_and_watch_extrinsic_until_without_events`.
- renamed example `event_callback` to `subscribe_events`. I believe that's more descriptive.
- Example `event_error_details`
     - renamed to `check_extrinsic_events`
     - added an extrinsic submission (that should succeed) to the example, which shows all the parameters that `ExtrinsicReport` contains with an descriptive `println` message. It also checks the returned events.

## Tests
- Updated `author_tests` to check for:
    - Returned `TransactionStatus`
    - That events are `None` or `Some` as one would expected. I therefore also changed from `submit_and_watch_extrinsic_until_without_events` to `submit_and_watch_extrinsic_until`. To still test `submit_and_watch_extrinsic_until_without_events` I added two new tests further below.
     
     
closes #591


examples/examples/check_extrinsic_events.rs
```diff
@@ -0,0 +1,142 @@
...
+
+	let alice = AccountKeyring::Alice.to_account_id();
+	let balance_of_alice = api.get_account_data(&alice).unwrap().unwrap().free;
+	println!("[+] Alice's Free Balance is {balance_of_alice}\n");
+
+	let bob = AccountKeyring::Bob.to_account_id();
+	let balance_of_bob = api.get_account_data(&bob).unwrap().unwrap_or_default().free;
+	println!("[+] Bob's Free Balance is {balance_of_bob}\n");
+
+	// First we want to see the events of a failed extrinsic.
```
echevrier, line 50: How do you want to do it?

echevrier (commented):
Really clean.  But I'm still missing the 'little' differences between `submit_and_watch_extrinsic_until_without_event` and `submit_and_watch_extrinsic_until`. 
- `submit_and_watch_extrinsic_until_without_event` doesn't thrown an error when the extrinsic has failed onchain. 
- how can we get the triggered events when the extrinsic has failed?

examples/examples/check_extrinsic_events.rs
haerdib: Good Input - let me do something about that

haerdib: https://github.com/scs/substrate-api-client/issues/617

I think that should be done in a different PR, as it affects the API. What do you think? 


Regarding your point of
> submit_and_watch_extrinsic_until_without_event doesn't thrown an error when the extrinsic has failed onchain.
What do you propose for this? I think the comments are quite telling already.. so I removed the `without_events` from the examples. It should not be used if one does not know what one is doing.

echevrier: https://github.com/scs/substrate-api-client/issues/617 makes sense. Adding the report to the Error gives access to all the information we have collected.

echevrier: > submit_and_watch_extrinsic_until_without_event doesn't thrown an error when the extrinsic has failed onchain.
> What do you propose for this? I think the comments are quite telling already.. so I removed the `without_events` from the examples. It should not be used if one does not know what one is doing.

Not showing any examples is just a way of forcing the user not to use it. Is that really what we want? 
I'm not sure the comments are clear enough. Because we're talking about an error when an extrinsic has failed on the chain, and not an event, the user may think that `submit_and_watch_extrinsic_until_without_event` informs her when an extrinsic fails on the chain. Without studying the code, they may miss this point. 
Maybe add an example `check_extrinsic_without_events` similar to `check_extrinsic_events`? We could do it in another issue. Maybe I'm too picky, so it's up to you to decide.

echevrier (approved):

examples/examples/check_extrinsic_events.rs
haerdib: Created a  follow up issue in https://github.com/scs/substrate-api-client/issues/627

Issue #624, opened 2023-07-20
got Error ExtrinsicNotFound although extrinsic succeeded (for long extrinsic)


brenzi:
If an extrinsic is very long, the `submit_and_watch_extrinsic_until_success` fails:

```
[...]

PR #636, opened 2023-07-28
Add unit tests for hash matching for unchecked extrinsic

haerdib:
Adds unit tests to ensure the calculated extrinsic hash matches the one from Substrate.

related to #624


echevrier (approved):
Thanks

PR #642, opened 2023-08-07
Increase timeout value

Niederb:
I ran the test multiple times and it seems to pass more reliably now


haerdib (approved):
Awesome, thanks!

PR #647, opened 2023-09-15
Polkadot v0.9.42 tag v0.14.0

clangenb:
Ported v0.14.0 to v0.9.42. 

Notable changes compared to polkadot-v1.0.0:
* Had to re-introduce the redefinitions of the substrate types because commit that introduced the `serde` feature flag is not yet in polkadot-v0.9.42
* some frame-system rpc's are not available in no-std.


haerdib:
It looks good - just be careful when using this branch. The no-std build does not seem to work completely yet.

PR #648, opened 2023-09-15
Add trait bounds to config: `Index: Default`, `BlockNumber: AtLeast32BitsUnsigned`

### LINKED ISSUE #646: `Config::Index` type should have default trait bound
https://github.com/scs/substrate-api-client/blob/dd6ca5ed2be71b177b0986e01cc0762655ad7d78/primitives/src/config/mod.rs#L37

clangenb:
Seems like reasonable assumptions for me, and they made my code in the worker a bit more ergonomic. :)

* Closes #646


haerdib (approved):
That makes sense to me. Thanks for helping out!

PR #657, opened 2023-10-25
Update frame metadata to v16

### LINKED ISSUE #608: Update dependencies and frame-metadata feature flags
After a `cargo update` we will need to change the feature flags of `frame-metadata`, since they removed the old ones.

see https://github.com/paritytech/frame-metadata/pull/62

Thanks @OverOrion for the ping in https://github.com/integritee-network/worker/issues/1388

Tasks:

- [ ] Update to frame-metadata v16.0.0
- [ ] Use `current` feature flag. This should already include metadata v15

haerdib:
Updates the frame-metadata to v16.0

To ensure the tests still run with the v14 -> v15 conversion, it was necessary to exchange the outdated ksm metdata from parity (older than 2 years) with a newer one, as it did not contain the enum "RuntimeCall" yet (see parity discussion: https://github.com/paritytech/subxt/issues/1171). To retrieve the metadata from a Kusama node, the file `dump_metadata` was created. This also allows fetching different metadata for testing purposes when needed.

Additionally, the following was updated:
- Updated rust-toolchain to a slightly newer one
- Fixed "tungstenite deprecated" clippy warnings
- Renamed `InvalidMetadataError`  to `MetadataConversionError` to clarify the error name. Also moved the Error Types from `MetadataError` to `MetadataConversionError` which were only used in metadata conversion.

closes #608


haerdib:
> Just a comment on the tests: you test against a correct .bin file and on-the-fly generated erroneous metadata. Would there be an old or invalid .bin file you can also test against?

Yes, there would be the old .bin file, which does not contain any "RuntimeCall" (and similiar) enums. But I don't think this should be tested against, since this is actually rather undesired, but hardly preventable behaviour (see subxt discussion:https://github.com/paritytech/subxt/pull/1174#discussion_r1335777924) . I don't know how to get erroneous metadata.

Do you think the manually generated metadata is not sufficient?

masapr:
So, it panics, when you use the old one? What I meant was to test the error-case to know how it behaves in that case. That's all.

But you can really leave it as is.

haerdib:
No, it is a legitimate worry, as this does break the compatibility with very old node runtimes. It does not panic, but it returns an error `MetadataConversionError::TypeNameNotFound( .. ) ` (see the test: https://github.com/scs/substrate-api-client/pull/657/files#diff-7f07d2e92e3f41eabe45f4f95c26fda3f502c19e35d851ea39aa34c303d42993R424-R425) 

But I think this should be alright, since it allows using the new v15 metadata (and reasonably old v14) and we are not supporting old polkadot-branches anyway.

PR #662, opened 2023-11-03
Add subscription ID recognition to tungstenite and ws-rpc client

### LINKED ISSUE #626: Fix Warning `Expected subscription, but received an id response instead`
See comment: https://github.com/scs/substrate-api-client/issues/620#issuecomment-1643844636:

Example hangs if an error should be returned.

Steps to reproduce:
run integritee-node 1.0.36 (polkadot-0.9.42 deps)
```
wget https://github.com/integritee-network/integritee-node/releases/download/1.0.36/integritee-node-dev
chmod +x integritee-node-dev
./integritee-node-dev --dev 
```

run substrate api client example:
```
git checkout polkadot-v0.9.42
export RUST_LOG=warn
cargo run -p ac-examples --example transfer_with_ws_client
```

Just logs the warning and keeps on waiting for a response.
``` 
[2023-07-24T13:25:21Z DEBUG substrate_api_client::rpc::tungstenite_client::client] got on_subscription_msg {"jsonrpc":"2.0","error":{"code":1002,"message":"Verification Error: Runtime error: Execution failed: Runtime panicked: Bad input data provided to validate_transaction: Could not decode `RuntimeCall::System.0`:\n\tCould not decode `Call`, variant doesn't exist\n","data":"RuntimeApi(\"Execution failed: Runtime panicked: Bad input data provided to validate_transaction: Could not decode `RuntimeCall::System.0`:\\n\\tCould not decode `Call`, variant doesn't exist\\n\")"},"id":"1"}
[2023-07-24T13:25:21Z WARN  substrate_api_client::rpc::tungstenite_client::client] Expected subscription, but received an id response instead: Object {"error": Object {"code": Number(1002), "data": String("RuntimeApi(\"Execution failed: Runtime panicked: Bad input data provided to validate_transaction: Could not decode `RuntimeCall::System.0`:\\n\\tCould not decode `Call`, variant doesn't exist\\n\")"), "message": String("Verification Error: Runtime error: Execution failed: Runtime panicked: Bad input data provided to validate_transaction: Could not decode `RuntimeCall::System.0`:\n\tCould not decode `Call`, variant doesn't exist\n")}, "id": String("1"), "jsonrpc": String("2.0")}
```

To be looked into: Is this a problem of ws-client/ tungstenite client only, but also jsonrpsee?

haerdib:
Updates the ws-rpc and tungstenite clients with the following features:
- Removal of the warning message "Expected subscription, but received an ID response instead: Object"
- The immediate subscription response (either the subscription ID or an error) is now interpreted by the client. All the following subscription messages are checked by the client if the ID matches, and only then are forwarded to the client. This now removes the need to ignore the "id" response that previously generated the responses, and does not allow mix-up of different subscription messages.
- adds tests to the CI to test the following three cases:
     -  Failed onchain extrinsic (Extrinsic went through successfully, only the Events indicate a failure of the xt)
     - Failed extrinsic before block submission (E.g. no subscription ID is returned but an immediate error response)
     - Successful extrinsic

closes #626

### DIFF
--- a/src/rpc/helpers.rs
+++ b/src/rpc/helpers.rs
@@ -1,0 +1,144 @@
+/*
+   Copyright 2019 Supercomputing Systems AG
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+	   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+*/
+
+use serde_json::Value;
+
+pub fn read_subscription_id(value: &Value) -> Option<String> {
+	value["result"].as_str().map(|str| str.to_string())
+}
+
+pub fn read_error_message(value: &Value, msg: &str) -> String {
+	match value["error"].as_str() {
+		Some(error_message) => error_message.to_string(),
+		None => format!("Unexpected Response: {}", msg),
+	}
+}
+
+pub fn subscription_id_matches(value: &Value, subscription_id: &str) -> bool {
+	match value["params"]["subscription"].as_str() {
+		Some(retrieved_subscription_id) => subscription_id == retrieved_subscription_id,
+		None => false,
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use serde_json::json;
+
+	#[test]
+	fn read_valid_subscription_response() {
+		let subcription_id = "tejkataa12124a";
+		let value = json!({
+			"result": subcription_id,
+			"id": 43,
+			"and_so_on": "test",
+		});
+
+		let maybe_subcription_id = read_subscription_id(&value);
+		assert_eq!(maybe_subcription_id, Some(subcription_id.to_string()));
+	}
+
+	#[test]
+	fn read_invalid_subscription_response() {
+		let subcription_id = "tejkataa12124a";
+		let value = json!({
+			"error": subcription_id,
+			"id": 43,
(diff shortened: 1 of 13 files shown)


src/rpc/tungstenite_client/client.rs
```diff
@@ -134,9 +134,25 @@ fn subscribe_to_server(
 	// Subscribe to server
 	socket.send(Message::Text(json_req))?;
 
+	// Read the first message response - must be the subscription id.
+	let msg = read_until_text_message(&mut socket)?;
+	let value: Value = serde_json::from_str(&msg)?;
```
masapr, line 139: As far as I understand, all the following logic is analyzing a string. I would put this logic into its own function and write a test for each flow through. Then you know, that it's doing what you expect.

masapr (approved):
Just one comment. But overall looks good to me.

src/rpc/tungstenite_client/client.rs
haerdib: I agree, unit testing does make sense here. But not just here, but also in the ws_client and other functions of this client. Would you agree to do that in a separate issue and PR?

masapr: I would add tests for new logic from this PR in this PR. For adding tests to existing code, I would create a new issue.

haerdib: Added issue: https://github.com/scs/substrate-api-client/issues/663, commit adding unit tests for changed code is incoming.

src/rpc/helpers.rs
```diff
@@ -0,0 +1,148 @@
+/*
```
masapr, line 1: I would've left this code (together with the tests) in the client.rs, as you're only using it there. It also would have been a simpler change.

masapr (approved):

src/rpc/helpers.rs
haerdib: Since both, ws and tungstenite client, are using these functions, I extracted it to the helpers. That way, I was able to remove duplicate code.

PR #676, opened 2023-11-22
Test previous tungstenite version

haerdib:
lets see what CI does


haerdib:
Closing this, because race condition was not the problem of #624

PR #677, opened 2023-11-27
Fix compile errors of stand-alone features

### LINKED ISSUE #644: `cargo check --no-default-features --features std,tungstenite` results in compiler error
Not sure if this is combination is intended to be used, but I think it should:

```
src/api/api_client.rs:183:51
    |
183 |         let (genesis_hash, metadata, runtime_version) = futures::future::try_join3(
    |                                                         ^^^^^^^ use of undeclared crate or module `futures`

error[E0433]: failed to resolve: use of undeclared crate or module `futures`
   --> src/api/api_client.rs:243:4
    |
243 |             futures::future::try_join(metadata_future, runtime_version_future).await?;
    |             ^^^^^^^ use of undeclared crate or module `futures`

```

Observed on https://github.com/scs/substrate-api-client/tree/polkadot-v0.9.42-tag-v0.14.0

haerdib:
Fixes the compile errors of standalone features. The following was changed:
- Added additional CI checks 
- Added `futures-util` to default import because the async code needs it to compile. Because the sync code does not, I only added the minimum needed package (which is -util and not the full `futures`).
- `tungstenite-client` and `ws-client` now include the `sync-api` feature, because they are only work in sync mode anyway.
- fixed async no-std compiliation by adding ```#[cfg(not(feature = "sync-api"))] use alloc::boxed::Box; ``` : async-trait automatically changes async return values into boxed values.

closes #644

### DIFF
--- a/node-api/src/metadata/metadata_types.rs
+++ b/node-api/src/metadata/metadata_types.rs
@@ -49,13 +49,6 @@ pub struct Metadata {
 }
 
 impl Metadata {
-	/// Access a pallet given its name.
-	#[deprecated(note = "please use `pallet_by_name` or `pallet_by_name_err` instead")]
-	pub fn pallet(&self, pallet_name: &str) -> Result<PalletMetadata<'_>, MetadataError> {
-		self.pallet_by_name(pallet_name)
-			.ok_or_else(|| MetadataError::PalletNameNotFound(pallet_name.to_string()))
-	}
-
 	/// An iterator over all of the available pallets.
 	pub fn pallets(&self) -> impl Iterator<Item = PalletMetadata<'_>> {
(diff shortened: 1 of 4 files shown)


Niederb (commented):
I don't quite understand why the changes with `Box` are needed. I assume you had compilation errors. Can you tell me how I can reproduce them?
```
#[cfg(not(feature = "sync-api"))]
use alloc::boxed::Box;
```

haerdib:
See the PR description:  
> - Fixed async no-std compiliation by adding `#[cfg(not(feature = "sync-api"))] use alloc::boxed::Box;` : async-trait automatically changes async return values into boxed values.

For reproducing: Simply remove the `alloc::boxed::Box;` and build run `cargo check --no-default-features`

Niederb (approved):
Now I understood. Thanks

PR #686, opened 2023-12-06
Update outdated crate dependencies

### LINKED ISSUE #671: Update outdated dependencies in cargo.tomls
There are some outdated dependencies in our .tomls that should be updated.

Example:
grafik

haerdib:
closes #671


haerdib:
Sending more than 58 requests seems to be a problem with the newer jsonrpsee.. I did not find out why as of now. Upstream issue: https://github.com/paritytech/jsonrpsee/issues/1259

PR #687, opened 2023-12-07
Update parity scale depenciens

haerdib:
part of #671


node-api/src/events/mod.rs
```diff
@@ -171,25 +177,7 @@ mod tests {
...
-
-		// Make sure that the bytes handed back line up with the fields handed back;
-		// encode the fields back into bytes and they should be equal.
-		let actual_fields = actual.field_values().expect("can decode field values (1)");
-		let mut actual_bytes = vec![];
-		for field in actual_fields.into_values() {
-			crate::scale_value::encode_as_type(&field, field.context, types, &mut actual_bytes)
-				.expect("should be able to encode properly");
-		}
-		assert_eq!(actual_bytes, actual.field_bytes());
```
haerdib, line 191: This assert failed after upgrading the scale libraries and was removed in subxt as well. I simply followed suit, didn't look  further into it. Since all the other tests are passing, I'm inclined to believe this is alright. Either way, there is a follow up issue: https://github.com/scs/substrate-api-client/issues/688

masapr (approved):
LGTM 👍

PR #691, opened 2023-12-07
Test jsonrpsee v0.20

haerdib:
Test PR for #692

### DIFF
--- a/src/rpc/jsonrpsee_client/mod.rs
+++ b/src/rpc/jsonrpsee_client/mod.rs
@@ -14,7 +14,7 @@
 use crate::rpc::{Error, Request, Result, RpcParams, Subscribe};
 use futures::executor::block_on;
 use jsonrpsee::{
-	client_transport::ws::{Uri, WsTransportClientBuilder},
+	client_transport::ws::{Url, WsTransportClientBuilder},
 	core::{
 		client::{Client, ClientBuilder, ClientT, SubscriptionClientT},
 		traits::ToRpcParams,
@@ -44,7 +44,7 @@ impl JsonrpseeClient {
 	}
 
 	pub async fn async_new(url: &str) -> Result<Self> {
-		let uri: Uri = url.parse().map_err(|e| Error::Client(Box::new(e)))?;
+		let uri: Url = url.parse().map_err(|e| Error::Client(Box::new(e)))?;
 		let (tx, rx) = WsTransportClientBuilder::default()
 			.build(uri)
 			.await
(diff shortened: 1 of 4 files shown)


src/rpc/jsonrpsee_client/mod.rs
```diff
@@ -64,11 +65,29 @@ impl Request for JsonrpseeClient {
...
+		let method_string = method.to_string();
+
+		// The inner jsonrpsee client must not deserialize to the `R` value, because the return value must
+		// implement `Send`. But we do not want to enforce the `R` value to implement this solely because we need
+		// to de-async something. Therefore, the deserialization must happen outside of the newly spawned thread.
+		// We need to spawn a new thread because tokio does not allow the blocking of an asynchronous thread:
+		// ERROR: Cannot block the current thread from within a runtime.
+		// This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
+		let string_answer: Value = std::thread::spawn(move || {
+			handle.block_on(client.request(&method_string, RpcParamsWrapper(params)))
```
niklasad1, line 83: You can't use block_on on tokio multi-thread it will panic.

The best would would be to change this to:

```rust
#[maybe_async::sync_impl]
impl Request for JsonrpseeClient {
	fn request<serde_json::Value>(&self, method: &str, params: RpcParams) -> Result<R> {
                let (tx, rx) = tokio::sync::oneshot();
 
                tokio::spawn(async move {
                    let res = self.inner.request(method, RpcParamsWrapper(params))).await.map_err(|e| Error::Client(Box::new(e)));
                    tx.send(res);
                });

                let rx = rx.blocking_recv().unwrap();
		let deserialized_value: R = serde_json::from_value(rx)?;
		Ok(deserialized_value)
	}
}
```

Otherwise, you need to create a single threaded tokio-runtime https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.new_current_thread

haerdib: The problem with the solution you're providing is, that it enforces the creation of a tokio task / thread outside of the request function, therefore breaking our current sync-api. With the current examples, it runs into the following runtime error at the `blocking_recv()` part:
```
ERROR: Cannot block the current thread from within a runtime.
This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```
I'm not entirely sure what we should do here - either provide an all-inclusive function with a handle `try_current` matching to whatever the user does outside, or enforce either tokio or non-tokio threading.

In the long run, probably the best thing would be to either remove sync-api altogether on our side or introduce a sync-api for jsonrspee without tokio enforcement.

niklasad1: ok, I see sorry for giving a bad suggestion. That's was news to me

haerdib: Not a bad suggestion at all - it really helped me along the way. Thank you very much!

haerdib:
closing this as obsolete - we removed the sync api as of #699

PR #698, opened 2023-12-19
Update WithExtrinsicParams to use ExtrinsicSigner<Self>

### LINKED ISSUE #628: `DefaultRuntimeConfig` in ExtrinsicSigner::<..> does not work
When initializing the Api with `DefaultRuntimeConfig` and a signer as following:
```rust
let mut api = Api::<DefaultRuntimeConfig, _>::new(client).unwrap();
api.set_signer(GenericExtrinsicSigner::<DefaultRuntimeConfig>::new(sudoer));
```

It does not compile due to:
```
note: expected struct `ExtrinsicSigner<AssetRuntimeConfig>`
   found struct `ExtrinsicSigner<WithExtrinsicParams<AssetRuntimeConfig, GenericExtrinsicParams<AssetRuntimeConfig, PlainTip<u128>>>>`
label: expected `ExtrinsicSigner<AssetRuntimeConfig>`, found `ExtrinsicSigner<WithExtrinsicParams<..., ...>>`
```

I suppose the error comes from the definition of `DefaultRuntimeConfig`:
```rust
/// Standard runtime config for Substrate and Polkadot nodes.
pub type DefaultRuntimeConfig =
	WithExtrinsicParams<AssetRuntimeConfig, PlainTipExtrinsicParams<AssetRuntimeConfig>>;
```

Workaround:
```rust
api.set_signer(GenericExtrinsicSigner::<_>::new(sudoer));
```
But that is still not nice. Isn't there a better way to define the `ExtrinsicSigner`?

haerdib:
- Updates `WithExtrinsicParams` to use the `ExtrinsicSigner<Self>` instead of the `T::ExtrinsicSigner`, which is defined as `ExtrinsicSigner<Self>` but with the other runtime.
- Simplifies examples to use `into()` instead of more verbose `ExtrinsicSigner::new`
- Use ExtrinsicSigner with `DefaultRuntimeConfig` on `chain_tests.rs`. Reasoning: Does not need a real signer, but it still ensures that `DefaultRuntimeConfig` is tested atleast as a Getter.

Fixes #628

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,8 +69,8 @@ dependencies = [
  "sp-application-crypto",
  "sp-core",
  "sp-runtime",
- "sp-runtime-interface",
- "sp-storage",
+ "sp-runtime-interface 17.0.0 (git+https://github.com/paritytech/polkadot-sdk.git?branch=master)",
+ "sp-storage 13.0.0 (git+https://github.com/paritytech/polkadot-sdk.git?branch=master)",
  "test-case",
 ]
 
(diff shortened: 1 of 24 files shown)


masapr (approved):
LGTM 👍

PR #699, opened 2023-12-20
Remove sync api jsonrpsee

### LINKED ISSUE #692: Remove future `block_on` to update to jsonrpsee v17+
While trying to update the jsonrpsee client to a newer version (#671), the client (most likely) deadlocked itself after 59 requests (see https://github.com/scs/substrate-api-client/actions/runs/7126321622/job/19404332411). This is because we are using `future::executor::block_on` within tokio runtime to make async code synchronous. More info: https://github.com/paritytech/jsonrpsee/issues/1259

To ensure this deadlock does not happen, I currently see two options:
1. Replace futures with a tokio block. 
Example replacement:
Old code:
```rust
impl Request for JsonrpseeClient {
	fn request<R: DeserializeOwned>(&self, method: &str, params: RpcParams) -> Result<R> {
		block_on(self.inner.request(method, RpcParamsWrapper(params)))
			.map_err(|e| Error::Client(Box::new(e)))
	}
}
```

New code (didn't find a better way yet, should be improved):
```rust
impl Request for JsonrpseeClient {
	fn request<R: DeserializeOwned>(&self, method: &str, params: RpcParams) -> Result<R> {
                // Catch the current runtime.
		let handle = match Handle::try_current() {
			Ok(handle) => handle,
			Err(_) => {
				// We are not inside a tokio runtime, so lets start one.
				let rt =
					tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
				rt.handle().clone()
			},
		};

		let client = self.inner.clone();
		let method_string = method.to_string();

		// The inner jsonrpsee client must not deserialize to the `R` value, because the return value must
		// implement `Send`. But we do not want to enforce the `R` value to implement this solely because we need
		// to de-async something. Therefore, the deserialization must happen outside of the newly spawned thread.
		// We need to spawn a new thread because tokio does not allow the blocking of the main thread:
		// ERROR: Cannot block the current thread from within a runtime.
		// This happens because a function attempted to block the current thread while the thread is being used to drive asynchronous tasks.
		let string_answer: Value = std::thread::spawn(move || {
			handle.block_on(client.request(&method_string, RpcParamsWrapper(params)))
		})
		.join()
		.unwrap()
		.map_err(|e| Error::Client(Box::new(e)))?;

		let deserialized_value: R = serde_json::from_value(string_answer)?;
		Ok(deserialized_value)
	}
}
```
2. Open PR on jsonrpsee providing sync api

haerdib:
As decided in #692, this PR removes the sync interface for the jsonrpsee client. Because now not all feauters are compatible with each other anymore (jsonrpsee may not be built with sync interface), the examples were split up between sync and async. This allows to distinguish the build and test them separately.

The following has been changed:
- Split up of examples in async and sync folders / versions
        - changed jsonrpsee examples to purely async ones
        - moved ws und tungstentite examples to the sync folder since they officially only support sync.
        - removed futures from jsonrpsee because tokio should be used instead for consistency
- Adapted CI to build the features separately and distinguish between building examples and unit testing.
- Removed the documentation part of the custom client example, because it was not easily possible to make it compile for both sync and async features and is already described in the README https://github.com/scs/substrate-api-client#rpc-client
- Unit tests now need a `block_on` because the default version is async. Since the api behind it is a dummy client, block_on should be fine.

closes  #692


node-api/src/error/mod.rs
```diff
@@ -8,7 +8,7 @@
 
 //! General node-api Error implementation.
 
-use alloc::{string::String, vec::Vec};
+use alloc::{boxed::Box, string::String, vec::Vec};
```
haerdib, line 11: Clippy warning - error was too long for being unconstrained.

src/api/api_client.rs
```diff
@@ -30,58 +30,6 @@ use sp_version::RuntimeVersion;
 /// Api to talk with substrate-nodes
 ///
 /// It is generic over the `Request` trait, so you can use any rpc-backend you like.
-///
-/// # Custom Client Example
```
haerdib, line 34: Removed the custom client example, because it was not easily possible to make it compile for both sync and async features and is already described in the README https://github.com/scs/substrate-api-client#rpc-client

masapr (approved):
LGTM 👍

PR #704, opened 2024-02-01
Add Encode Decode to ExtrsinicReport and use RawExtrinsicDetails

### LINKED ISSUE #397: Add Encode / Decode to ExtrinsicReport
With PR https://github.com/scs/substrate-api-client/pull/386 a new struct `ExtrinsicReport` has been introduced. For a good user experience, it should implement at least Encode and Decode, maybe even Serialize and Deserialize. 

However, because `ExtrinsicReport` contains `EventDetails`, which do not yet support Encode/Decode, it first needs to be investigated how to fix this. Or if it's not fixable, maybe skip the EventDetails during Serilaization with serde::skip or something alike.

See https://github.com/scs/substrate-api-client/pull/386#discussion_r1055177318

haerdib:
- Removed Pointer and usize from `EventDetails` as they do not implement codec Decoding / Encoding.
- Added `RawEventDetails` to allow usage of EventDetails without always holding the full metadata
- Changed `ExtrinsicReport` to use `RawEventDetails` instead, to make the report smaller. Otherwise, the full metadata would get encoded / decode and moved all the time. I'd rather keep the report small. If needed, one can always transform the `RawEventDetails` back to `EventDetails`

closes  #397

### DIFF
--- a/node-api/src/events/raw_event_details.rs
+++ b/node-api/src/events/raw_event_details.rs
@@ -28,7 +28,7 @@ use scale_value::{scale::TypeId, Composite};
 /// Raw event details without the associated metadata
 // Based on subxt EventDetails.
 // https://github.com/paritytech/subxt/blob/8413c4d2dd625335b9200dc2289670accdf3391a/subxt/src/events/events_type.rs#L197-L216
-#[derive(Debug, Clone, Encode, Decode)]
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
 pub struct RawEventDetails<Hash: Decode + Encode> {
 	phase: Phase,
 	/// The index of the event in the list of events in a given block.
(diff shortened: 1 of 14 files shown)


masapr (changes requested):
Since the reason for this change, is that we want to support serialization/deserialization, I would add a test for that. Then we know this capability will not be removed without a failing test.

Apart from that it looks good to me

src/api/mod.rs
```diff
@@ -301,4 +301,18 @@ mod tests {
 		assert!(TransactionStatus::Dropped.reached_status(status));
 		assert!(TransactionStatus::Invalid.reached_status(status));
 	}
+
```
masapr, line 305: This is the test I meant 👍

node-api/src/events/raw_event_details.rs
```diff
@@ -326,3 +326,535 @@ impl<Hash: Encode + Decode> RawEventDetails<Hash> {
 		Ok(decoded)
 	}
 }
+
```
masapr, line 329: hmm, not sure if we need all these tests. What I meant was just the high level one you added below ...

haerdib: I know. But since I was at it already and needed to create new EventDetails either way, I copied & adapted the other unit tests. Is that not acceptable?

masapr: No, it's ok now

masapr (approved):
After our discussion, I'd say it's good now

PR #706, opened 2024-02-05
Split compose_call into two functions

### LINKED ISSUE #598: Split `compose_call` into  two functions
The current compose_call! macro expects the full metadata and the pallet name as input. But in case multiple calls for one pallet are created, the current code is forced to get the `PalletMetada` each time for each call. That's not necessary.

Therefore I'm proposing to split it into two functions:
```rust
/// Generates the extrinsic's call field for a given module and call passed as &str
/// # Arguments
///
/// * 'node_metadata' - This crate's parsed node metadata as field of the API.
/// * 'pallet' - Pallet name as &str for which the call is composed.
/// * 'call_name' - Call name as &str
/// * 'args' - Optional sequence of arguments of the call. They are not checked against the metadata.
/// As of now the user needs to check himself that the correct arguments are supplied.
#[macro_export]
macro_rules! compose_call {
($node_metadata: expr, $pallet: expr, $call_name: expr $(, $args: expr) *) => {
        {
            let pallet_metadata = $node_metadata.pallet_by_name($pallet).unwrap();
            $crate::compose_call_for_pallet_metadata !(pallet_metadata, $call $(, ($args)) *);
        }
    };
}

/// Generates the extrinsic's call field for the given PalletMetadata
/// # Arguments
///
/// * 'pallet_metadata' - This crate's parsed pallet metadata as field of the API.
/// * 'call_name' - Call name as &str
/// * 'args' - Optional sequence of arguments of the call. They are not checked against the metadata.
/// As of now the user needs to check himself that the correct arguments are supplied.
#[macro_export]
macro_rules! compose_call_for_pallet_metadata {
($pallet_metadata: expr, $call_name: expr $(, $args: expr) *) => {
        {
            let call_index = pallet.call_variant_by_name($call_name).unwrap().index;
            ([pallet.index(), call_index as u8] $(, ($args)) *)
        }
    };
}

```

haerdib:
closes #598

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,9 +16,12 @@ dependencies = [
 name = "ac-compose-macros"
 version = "0.4.2"
 dependencies = [
+ "ac-node-api",
  "ac-primitives",
+ "frame-metadata 16.0.0",
  "log",
  "maybe-async",
+ "parity-scale-codec",
 ]
 
 [[package]]
--- a/compose-macros/Cargo.toml
+++ b/compose-macros/Cargo.toml
@@ -17,6 +17,12 @@ maybe-async = { version = "0.2.7" }
 # local
 ac-primitives = { path = "../primitives", default-features = false }
 
+[dev-dependencies]
+ac-node-api = { path = "../node-api" }
+frame-metadata = { version = "16.0" }
+codec = { package = "parity-scale-codec", version = "3.6.1" }
+
+
 [features]
 default = ["std", "sync-api"]
 sync-api = ["maybe-async/is_sync"]
--- a/compose-macros/src/lib.rs
+++ b/compose-macros/src/lib.rs
@@ -37,11 +37,25 @@ mod rpc;
 macro_rules! compose_call {
 ($node_metadata: expr, $pallet: expr, $call_name: expr $(, $args: expr) *) => {
         {
-            let pallet = $node_metadata.pallet_by_name($pallet).unwrap().to_owned();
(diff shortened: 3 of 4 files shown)


masapr (changes requested):
In the issue you say "in case multiple calls for one pallet are created, the current code is forced to get the PalletMetada each time for each call. That's not necessary."

I don't see, how you can create this scenario with the new macros. Would you call 
`let pallet_metadata = $node_metadata.pallet_by_name($pallet).unwrap().to_owned();`
directly and then use the second macro to compose the calls?

Generally, I would propose to write a simple test, with the scenario described in #598. If it can be done nicely with the new code, then I'm fine with it.

haerdib:
Added an unit test in [688068a](https://github.com/scs/substrate-api-client/pull/706/commits/688068a16151507d421a880889ae96e8c6776a23). Is that what you imagined?

masapr:
> Added an unit test in [688068a](https://github.com/scs/substrate-api-client/pull/706/commits/688068a16151507d421a880889ae96e8c6776a23). Is that what you imagined?

Yes. But I would also add some assertion on _call_one and _call_two.
... sorry for being so picky here. I know our code-base is not really written with TDD. But if we specifically change something for a specific purpose, why not check, that what we aim for, actually works.

haerdib:
Cool! You're right - I fixed it in [5de32cf](https://github.com/scs/substrate-api-client/pull/706/commits/5de32cf352ddbcd4e3bddefd7e28869fe9dba3d9) and moved it to the correct crate as well.

masapr (approved):
LGTM, thanks for the test :-)

PR #707, opened 2024-02-05
Update jsonrpsee to v0.21

### LINKED ISSUE #702: Update jsonrpsee library to v20+
blocked by #692

haerdib:
closes #702

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ serde_json = { version = "1.0.79", default-features = false }
 url = { version = "2.0.0", optional = true }
 
 # websocket dependent features
-jsonrpsee = { version = "0.16", optional = true, features = ["async-client", "client-ws-transport", "jsonrpsee-types"] }
+jsonrpsee = { version = "0.21", optional = true, features = ["async-client", "client-ws-transport-native-tls", "jsonrpsee-types"] }
 tungstenite = { version = "0.21", optional = true, features = ["native-tls"] }
 ws = { version = "0.9.2", optional = true, features = ["ssl"] }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -619,11 +619,13 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
 [[package]]
 name = "async-lock"
-version = "2.8.0"
+version = "3.3.0"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -42,7 +42,7 @@ serde_json = { version = "1.0.79", default-features = false }
 url = { version = "2.0.0", optional = true }
 
 # websocket dependent features
-jsonrpsee = { version = "0.16", optional = true, features = ["async-client", "client-ws-transport", "jsonrpsee-types"] }
+jsonrpsee = { version = "0.21", optional = true, features = ["async-client", "client-ws-transport-native-tls", "jsonrpsee-types"] }
```
masapr, line 45: This is less on this PR but a general question: It seems wrong to me, that we explicitly reference the different websocket libraries here. This does mean, that I have to include tungstenite, even when I'm using the api client with jsonrpsee. Is this correct? 

On the specific code here: What if I want to use the api client with jsonrpsee, but without the native-tls feature? As far as I understand, there's no reason to not allow this. But it seems to be prevented by this.

masapr (changes requested):

Cargo.toml
haerdib: >This does mean, that I have to include tungstenite, even when I'm using the api client with jsonrpsee. Is this correct?

No, that is not correct. The dependency imports of the rpc-clients is optional and only included if the correspondent feature is chosen.

> What if I want to use the api client with jsonrpsee, but without the native-tls feature? As far as I understand, there's no reason to not allow this. But it seems to be prevented by this.

Do you have something in mind? We could forward the features of the jsonrpsee client, but that would mean more features on our side, I believe… or is there a way around it?
Or we could allow that the user constructs his/her own jsonrpsee client and just wraps our wrapper struct around it: https://github.com/scs/substrate-api-client/blob/master/src/rpc/jsonrpsee_client/mod.rs#L30-L33:
```rust
// Something like this:
let jsonrpsee_client = ClientBuilder::default()
		.max_buffer_capacity_per_subscription(4096)
		.build_with_tokio(tx, rx);
let api_client = JsonrpseeClient { inner: Arc::new(client) };
```

haerdib: Actually, I'm not sure if my first approach would work as long as we are constructing our own jsonrpsee-client.

PR #717, opened 2024-02-15
CI: Ensure unit tests are run in all crates

haerdib:
With the `--all-features` the async version of the unit tests is not needed anymore.
Reasoning: Previously, `cargo test --release` has built the api-client in default mode, which is async. With the newly added `--workspace` the client is now built in sync mode. 

I believe that's reasonable, as we already test the async version thoroughly in the system tests.


masapr:
Can you give some more information? I don't understand this issue, especially not why the block_on are not needed anymore.

haerdib:
Updated the PR description. Does it make more sense now?

masapr (approved):
perfect, thanks for adjusting the description

Issue #720, opened 2024-02-15
Use new docker image of substrate in CI


masapr:
The CI gives warnings:
`Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`

After checking, I saw that we are still using an old docker image from parity for our tests:
https://github.com/scs/substrate-api-client/blob/839ba6f0c8fcda0efc0733f635056a91b6df4355/.github/workflows/ci.yml#L178

I'm not sure, which is the correct new one. So this should be checked and then the CI code updated

haerdib:
There does not seem to be an up-to-date Substrate docker atm: https://github.com/paritytech/polkadot-sdk/issues/2750

I wonder if it's worth the effort to self-build a node, or if we wait for parity to fix this?

masapr:
I would expect that Parity has some docker image, that let's you run a node, no? Or do you think they don't have that anymore? Are you sure it's not called differently or that there is another way of running a dummy substrate node?

haerdib:
If they do, I didn't find it in docker hub. And also I'd also expect them to update their doc accordingly, but they haven't: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/docker/run.sh#L11

So I guess the above linked issue, especially the comment https://github.com/paritytech/polkadot-sdk/issues/2750#issuecomment-1862636921 is the issue here.

masapr:
ok, I see. Thanks for checking. Then let's close this issue. Looks like there will be a new docker image eventually with the same name on dockerhub

haerdib:
Let's keep it so we don't forget about it. Maybe there will be a renaming or something.

Option: paritypr/substrate, which is used for zombienet. https://github.com/paritytech/polkadot-sdk/blob/5fc7622cb312f2d32ec8365012ee0a49622db8c8/substrate/zombienet/0001-basic-warp-sync/generate-warp-sync-database.toml#L4
This still seems to be updated. But no idea what kind of node is run there.

haerdib:
After discussion: We could try to use the paritypr image one. Lets see if CI accepts that.

masapr:
[...]

PR #721, opened 2024-02-19
CI: Use paritypr substrate node

haerdib:


masapr (approved):
Thanks for the change.
Unfortunately we still get the original warning
`Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: styfle/cancel-workflow-action@0.11.0. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`
So let's remove the "closes #720". Nevertheless, I think this change makes sense, so let's merge it.

haerdib:
I have created a separate issue instead, as this warning is likely generated by a different part of our CI: #723

PR #722, opened 2024-02-19
Return ExtrinsicReport also for failed extrinsic

### LINKED ISSUE #617: Return full extrinsic information, even when xt has failed
We currently throw away all xt information if it has failed: 

In function
https://github.com/scs/substrate-api-client/blob/47d1173c0299050555f2bbb6b7765837c049a031/src/api/rpc_api/author.rs#L309-L330
We only return the dispatch error if the xt has failed.  This only tells us why the extrinsic has failed, but not the xt_hash, nor the events for it. Even though it has been fetched previously.

So I'm proposing to return an Error that returns all the information: see https://github.com/scs/substrate-api-client/pull/616

haerdib:
closes #617


haerdib:
Replaced by #725

PR #725, opened 2024-02-19
Add failed extrinsic error

haerdib:
When an extrinsic fails onchain, the full encoded report is now returned. This allows the user to retrieve the associated events, blockhash and the dispatch error.


masapr (approved):
LGTM 👍

Issue #726, opened 2024-02-23
release api-client v0.17.0


haerdib:
- [x] update cargo.toml versions
- [x] add new tag
- [x] release new release text

haerdib:
Released: https://github.com/scs/substrate-api-client/releases/tag/v0.17.0

PR #715, opened 2024-02-14
Update rust-toolchain and dependencies

haerdib:
closes #712 
closes  #708

PR #729, opened 2024-02-23
Bump toml versions

haerdib:
Updates the following tomls:
- api-client to new, matching release v0.17.0
- client-keystore: breaking changes in https://github.com/scs/substrate-api-client/pull/715
- ac-compose-macros: breaking changes in https://github.com/scs/substrate-api-client/pull/709
- examples and testing: breaking changes due to splitting up the examples in async and sync crates
- ac-node-api: breaking changes in https://github.com/scs/substrate-api-client/pull/704
- ac-primitves: no breaking changes, but small updates in https://github.com/scs/substrate-api-client/pull/703

related to #726

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,7 +14,7 @@ dependencies = [
 
 [[package]]
 name = "ac-compose-macros"
-version = "0.4.2"
+version = "0.5.0"
 dependencies = [
  "ac-node-api",
  "ac-primitives",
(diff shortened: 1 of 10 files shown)


masapr (approved):
I'm not sure, it's good practice to have different versions for different packages. Couldn't we put all packages to 0.17.0 ?

haerdib:
We could, but how can we track breaking changes for the different crates then? AFAIK it is not common in rust to have one version for all crates.

PR #744, opened 2024-03-01
bend client-keystore to crates io for v0.17.0 release

brenzi:
CC @haerdib @masapr 

It would be great if you could publish the client-keystore to crates.io as well

Not sure if you want to rename it first to match the scheme: i.e. `ac-keystore`

_edit_
however, as client-keystore is copy-paste from the sdk I'll try to use the original instead....maybe this here isn't necessary

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,13 +59,13 @@ sp-version = { default-features = false, features = ["serde"], version = "31.0" 
 frame-support = { optional = true, version = "30.0" }
 
 # local deps
-ac-compose-macros = { version = "0.5", default-features = false }
-ac-node-api = { version = "0.6", default-features = false }
-ac-primitives = { version = "0.9", default-features = false }
+ac-compose-macros = { version = "0.5", path = "compose-macros", default-features = false }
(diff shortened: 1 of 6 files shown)


brenzi:
hmm, we'll need ac_node_api bumped on crates.io as well

Cargo.toml
```diff
@@ -59,13 +59,13 @@ sp-version = { default-features = false, features = ["serde"], version = "31.0"
 frame-support = { optional = true, version = "30.0" }
 
 # local deps
-ac-compose-macros = { version = "0.5", default-features = false }
-ac-node-api = { version = "0.6", default-features = false }
-ac-primitives = { version = "0.9", default-features = false }
+ac-compose-macros = { version = "0.5", path = "compose-macros", default-features = false }
```
brenzi, line 62: using local paths even if published to crates.io avoids duplicates when patching

Niederb:
@brenzi Can we close this issue? The keystore is now available on crates: https://crates.io/crates/ac-keystore

PR #746, opened 2024-03-02
make stable work for std build

### LINKED ISSUE #745: make stable build work for std
There is an upstream issue for the polkadot-sdk which makes us build our node in stable:
https://github.com/paritytech/polkadot-sdk/pull/3296

api client currently uses a nightly feature which is only necessary for no_std builds. I suggest to enable this feature only for no_std builds

PR will come shortly

CC #581

brenzi:
closes #745


masapr (approved):
Thanks for the PR @brenzi !

PR #750, opened 2024-03-12
Add the crates.io version numbers for local dependencies

Niederb:
It is allowed to have the local path as well as the version number in the dependencies. For local builds the path is used (but the version number is still verified) and on crates.io the version number is used. 
See also this description:
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations
This should make releases in the future simpler as we don't have to add the version numbers later. See for example this branch that adds the version numbers:
https://github.com/scs/substrate-api-client/pull/742


masapr (approved):
pretty cool. Thank you!

PR #752, opened 2024-03-14
Upgrade `mio` to fix security advisory

Niederb:
Just did a `cargo update`. This should fix the security advisory:
https://github.com/scs/substrate-api-client/security/dependabot/27


Niederb:
Closed in favor of #756

PR #754, opened 2024-03-14
Rename crate to ac-keystore

### LINKED ISSUE #716: Suggestion: Rename substrate-client-keystore to ac-keystore
All our crates are named with ac-.. , where ac is short for `api-client`, except for the substrate-client-keystore. 

ac-keystore was still free at time of writing this issue. Before doing it, one should check if the crate name is still available: https://crates.io/search?q=ac-keystore

Niederb:
To be more consistent with the remaining crates. I also renamed the directory to be more consistent with the rest

Closes #716


masapr (approved):
LGTM 👍

PR #755, opened 2024-03-14
Change dependencies to point to the crates.io versions

Niederb:
Only for release to crates.io. 
Do not merge!

### DIFF
--- a/keystore/Cargo.toml
+++ b/keystore/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ac-keystore"
-version = "0.10.0"
+version = "0.10.1"
 authors = ["Supercomputing Systems AG <info@scs.ch>"]
 license = "Apache-2.0"
 edition = "2021"
--- a/keystore/Cargo.toml
+++ b/keystore/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.10.0"
 authors = ["Supercomputing Systems AG <info@scs.ch>"]
 license = "Apache-2.0"
 edition = "2021"
+repository = "https://github.com/scs/substrate-api-client"
+description = "Keystore (and session key management) for ed25519 based chains like Polkadot. (fork of sc-keystore)"
 
 [dependencies]
 array-bytes = "6.2"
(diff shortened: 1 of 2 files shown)


Niederb:
@masapr Can you have a look at the meta-infos? I would then revert the dependency paths back to github and then we can actually merge this.

keystore/README.md
```diff
@@ -0,0 +1,3 @@
+# ac_keystore
+
+This crate is a submodule of the [substrate-api-client](https://github.com/scs/substrate-api-client). It contains a fork of the sc-keystore crate and this is in order to access private functions and members of the crate.
```
masapr, line 3: I would write the second sentence like this:

It is a fork of substrate's sc-keystore crate. The substrate-api-client needs its own fork in order to access private functions and members of the crate.

keystore/Cargo.toml
```diff
@@ -12,11 +15,11 @@ parking_lot = "0.12.0"
...
-sc-keystore = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "master" }
-sp-application-crypto = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "master" }
-sp-core = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "master" }
-sp-keyring = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "master" }
-sp-keystore = { git = "https://github.com/paritytech/polkadot-sdk.git", branch = "master" }
+sc-keystore = { version = "27.0.0" }
+sp-application-crypto = { version = "32.0.0" }
+sp-core = { version = "30.0.0" }
+sp-keyring = { version = "33.0.0" }
+sp-keystore = { version = "0.36.0" }
```
masapr, line 22: This part we only need for the publishing, right? But we want to update the text below in general, is that right?
then this should be two PRs? Resp. only the PR for the text. And the dependencies here would only be adjusted for the branch that is published on crates.io.

masapr (changes requested):

keystore/README.md
Niederb: I split the PR into two and added your suggestion to the new PR. See #757

keystore/Cargo.toml
Niederb: Opened a separate PR. See https://github.com/scs/substrate-api-client/pull/757

Niederb:
Changes are only needed to release to crates.io. Closing this

PR #756, opened 2024-03-14
Update mio dependency that has a vulnerability

Niederb:
This should fix the security advisory:
https://github.com/scs/substrate-api-client/security/dependabot/27


masapr (approved):
Not sure if it's still needed (see comment in link). But I don't see an issue with updating, so 👍

PR #771, opened 2024-05-22
Create a nightly build

Niederb:
Build should run once per day and also do a `cargo update` in order to make sure the api-client still builds against the latest polkadot master


masapr (approved):
👍

Issue #769, opened 2024-05-13
Investigate WebAssembly support for '-client' feature


Niederb:
Check if one of our supported client crates (jsonrpsee, tungstenite or ws) has support for WebAssembly or if there would be another way to implement this (using wasi or using another crate).

PR #772, opened 2024-05-28
Wasm test

Niederb:
Mostly scaffolding to have a WebAssembly test in the build. The test itself is not yet super meaningful, but I think it is still a good first step to have something running and tested.
For a proper test with real use-cases we need to investigate how to create outside connections. See #769

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,7 @@ jobs:
           nc -z -v 127.0.0.1 9944
           chmod +x ${{ matrix.example }}
           ./${{ matrix.example }}
+
   wasm_examples:
     runs-on: ${{ matrix.os }}
     needs: build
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -235,6 +244,31 @@ jobs:
           nc -z -v 127.0.0.1 9944
           chmod +x ${{ matrix.example }}
           ./${{ matrix.example }}
+  wasm_examples:
```
masapr, line 247: I would add a newline before this, or even put this block below the next block "merge", as the merge is using only the examples. I'm almost wondering if we should put the wasm-stuff into its own yml-file (?). But I leave this up to you

masapr (changes requested):
Thanks! Looks like a good first step. Only one minor comment on the yml.

.github/workflows/ci.yml
Niederb: I added the newline. The merge is actually also for the wasm-example as this can also be downloaded. I also updated the merge to reflect this.

masapr (approved):
LGTM

PR #774, opened 2024-06-04
Upload Cargo.lock during build

Niederb:
A small improvement to our build such that we also upload the Cargo.lock. This is mostly useful for the nightly build which updates the Cargo.lock but since it is a small file we can include it in all the builds.


masapr (approved):
👍

PR #775, opened 2024-06-11
Implement balance_transfer_keep_alive

Niederb:


testing/async/examples/pallet_balances_tests.rs
```diff
@@ -15,15 +15,56 @@
...
 
 #[tokio::main]
 async fn main() {
 	// Setup
 	let client = JsonrpseeClient::with_default_url().await.unwrap();
-	let api = Api::<AssetRuntimeConfig, _>::new(client).await.unwrap();
+	let mut api = Api::<AssetRuntimeConfig, _>::new(client).await.unwrap();
 
-	let _ed = api.get_existential_deposit().await.unwrap();
+	let ed = api.get_existential_deposit().await.unwrap();
```
masapr, line 30: I think the focus of this test is sending funds, therefore I would make a new test-file with a fitting name. Also I would test a successfull transaction with balance_transfer_keep_alive.

I would create 2 new tests: 

1. "send_funds_test.rs" that tests a successfull balance_transfer_keep_alive transaction
2. "send_funds_with_death_test.rs" for the test case here.

masapr (changes requested):

testing/async/examples/pallet_balances_tests.rs
Niederb: I agree with adding a successful transaction for `balance_transfer_keep_alive` but I would keep it all in the same file it is now. The reason is that so far the testing examples are more organized by pallet/subsystem (`frame_system_tests.rs`, `pallet_contract_tests.rs`, `chain_tests.rs`, ....) and not by a specific test case.

masapr (approved):
LGTM 👍

PR #776, opened 2024-06-13
Check metadata hash

Niederb:
Contains the minimal changes needed to support the new extrinsic format with metadata-hash but actually disables the check.

We should discuss how/if we want to support this kind of breaking change by supporting backwards compatibility. In theory we would need to run the whole testing pipeline for both versions and for some cases we would need separate test-data. Already in this PR I had to deactivate two tests as the data is not compatible anymore (and I have honestly no idea how the file `test-runtime.compact.wasm` was generated)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,7 @@ jobs:
 
                 # Test for 32 bit and wasm32-unknown-unknown compatibility
                 cargo build --target wasm32-unknown-unknown --no-default-features --features sync-api,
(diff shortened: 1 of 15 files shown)


masapr (approved):
LGTM
+ can you create a new issue for the discussion on backward compatibility?

Niederb:
> LGTM + can you create a new issue for the discussion on backward compatibility?

I created an issue regarding backwards compatiblity (#788) and one to re-activate the disabled tests (#789)

Issue #779, opened 2024-06-19
Incorporate stabilized APIs


Niederb:
The functions we test in `unstable_rpc_api_calls.rs` were stabilized:
https://github.com/paritytech/polkadot-sdk/issues/1156
https://github.com/paritytech/polkadot-sdk/issues/1508

We should provide these functions as actual API functions.

Niederb:
After looking a bit more into this it is my understanding that the stabilized calls are part of the new JSON-RPC and do not offer anything fundamentally new. So it does not make sense to incorporate the calls in the API. Instead we should think about how/when to migrate to the new JSON-RPC. See #764 for that.

PR #783, opened 2024-06-26
Use stabilized chainSpec calls in `unstable_rpc_api_calls`

Niederb:
This is a quick fix to get the test working again. I created a follow up issue to provide an actual API for the functionality in the `unstable_rpc_api_calls.rs` test. See https://github.com/scs/substrate-api-client/issues/779

In this PR:
- Use stabilized versions of chainSpec calls
- Use newer docker image (From 01.11.2023)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,7 +216,7 @@ jobs:
 
       - name: Run latest node
         run: |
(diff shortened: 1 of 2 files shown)


masapr (approved):
LGTM

PR #798, opened 2024-08-28
Transform "unstable" example into an "new JSON-RPC" example

Niederb:
The calls in the example are not unstable anymore so the name does not make sense anymore
- Change the file name as the calls are not unstable anymore
- Given that the calls are stable we don't need to verify that they exist
Also see comments in issue #779


masapr (approved):
👍

PR #802, opened 2024-09-11
Examples: switch to use rococo node

### LINKED ISSUE #778: Use substrate image, that is kept up-to-date
With the current image we use: paritypr/substrate:latest we have the issue, that it is not updated. We should change to parity/polkadot (if possible). This seems to be the official docker image and is currently updated frequently. Also note the current bugreport: #761

haerdib:
This PR updates the following:
- Switches CI testing node to polkadot docker image, which is regularly updated and uses `../latest`. The dev runtime is rococo.
- Added a new Runtime Config, which matches the rococo polkadot runtime:
```rust 
pub type RococoRuntimeConfig = WithAddress<DefaultRuntimeConfig, MultiAddress<AccountId32, ()>>;
```
- Default Runtime was switched to be the "original" defined one. This makes the generation of the `rococo_runtime_config` a little bit more comprehensible / traceable.
- The following tests were removed:
    - staking
    - contracts
    - identity
    Reasoning: Staking and contracts pallet are not available on the Rococo runtime. The Identity pallet is available, but the calls are filtered: https://github.com/paritytech/polkadot-sdk/blob/cc7ebe054038186c8a7409f54c4008ddc82a5ad0/polkadot/runtime/rococo/src/lib.rs#L189-L198
- Created a new wasm binary for the runtime upgrade: needed a higher spec version (must be higher than the original one) and a renaming of the runtime (needs to have the same name as the old one)

closes #778


examples/async/examples/minimal_template_runtime.compact.compressed.wasm
Niederb, line 1: Do we have documentation on how to get this file and what version this currently is?

testing/async/examples/pallet_contract_tests.rs
```diff
@@ -1,79 +0,0 @@
...
-async fn main() {
-	// Initialize api and set the signer (sender) that is used to sign the extrinsics.
-	let signer = AccountKeyring::Alice.pair();
-	let client = JsonrpseeClient::with_default_url().await.unwrap();
-	let mut api = Api::<AssetRuntimeConfig, _>::new(client).await.unwrap();
-	api.set_signer(signer.into());
-
-	println!("[+] Alice's Account Nonce is {}", api.get_nonce().await.unwrap());
-
-	let wasm = include_bytes!("flipper.wasm").to_vec();
```
Niederb, line 54: I think `flipper.wasm` is not used anymore

Niederb (commented):
I like the changes. I think it makes sense to use the rococo runtime as this one should be better maintained.

examples/async/examples/minimal_template_runtime.compact.compressed.wasm
haerdib: I built it myself. But effectively it can be any runtime, as long as the spec_version is higher than the original one and the runtime name matches the old one. I only took the minimal one to make sure it builds faster.

Should I add a documentation for this?

testing/async/examples/pallet_contract_tests.rs
haerdib: Good catch, thanks!

examples/async/examples/minimal_template_runtime.compact.compressed.wasm
Niederb: Would be great to have some documentation 👍

Niederb (approved):
Nice. Thanks for the updated documentation 👍

Issue #800, opened 2024-09-02
New release names in polkadot-sdk


Niederb:
`polkadot-sdk` change to a different naming schema for their releases:
https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2407

We should look into what this means for us. Should we also change the naming of our releases and branches?

haerdib:
Yes, we should definitely change our release name as well. 
Suggested branch name: `polkadot-stable2407-2` (exactly the same as the one Parity uses). What do you think?

As crates.io is now used by Parity and our users as well, we should probably add a crates.io release cycle as well. Are we doing something like that already?

Niederb:
I have to say I'm not a big fan of the naming approach of parity. I find it very unclear how stable2407, stable2407-1 and stable2407-02 are related (for example stable2407-2 was released in September). Still I think for the branches it makes sense to use the same as parity. 
I wonder if for crates.io a release is always tied to a polkadot release? And how does a user know which polkadot version is supported? 
Parity seems to write a version in the tagline of the crate which again does not make sense to me as this is not tied to a version: https://crates.io/crates/pallet-timestamp/

haerdib:
Idea from encointer: Use major version to indicate a new api-client release and minor for polkadot updates.

E.g. 1.2.0 -> version 1 from api-client and version 2 of polkdaot. If polkadot updates and we follow, it will be released as 1.3.0. 
If the api-client creates a new release it will be 2.0.0. Does this sound correct @Niederb ?

Niederb:
Yes, that's the general idea

haerdib:
Then a concrete TODO to solve this issue would be:
- [ ] update to newest parity release (on github master, whichever commits fits the newest release)
- [ ] publish a new crates.io release with the version 1.0.0
- [ ] create a new branch pointing to the respective release branch (is this still needed?)

Or should we wait with the version 1.0.0 for something specific?

Niederb:
If we adapt the versioning approach I would increase the major version to prevent/reduce confusion when someone is updating

Niederb:
See #809

PR #808, opened 2024-10-29
Create release for polkadot 2409

Niederb:
DO NOT MERGE
- Prepare release for polkadot 2409 (=1.16.0)
- Adapt version schema as discussed in #800
  - Decided to go for v1.16.0
    - Major version of 1 to signal changes in the versioning strategy
    - Minor version of 16 to represent polkadot 1.16.0
  - Also see #809 
- In the past we also provided a branch that pointed to a release branch on the `polkadot-sdk` Github respository but I suggest we focus on crates.io releases and only provide a branch that points to crates.io. 
- Build is still not passing due to compilation issues when we pull the versions from crates.io
  - Same problem as with previous releases
  - This is because the `solochain-template-runtime` dependency is not released on crates.io and this leads to all sorts of incompatible, duplicated types.
  - At least we are down to a single test not compiling
    - 5 failing tests in 0.17.0 https://github.com/scs/substrate-api-client/actions/runs/8099132229
    - 3 failing tests in 0.17.0 https://github.com/scs/substrate-api-client/actions/runs/9856778531

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,7 +14,7 @@ dependencies = [
 
 [[package]]
 name = "ac-compose-macros"
-version = "0.18.0"
+version = "1.16.0"
 dependencies = [
  "ac-node-api",
  "ac-primitives",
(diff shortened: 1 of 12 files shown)


masapr (approved):
After discussion on pros and cons on the new version numbering, I approve. I think this is the most practical way.

haerdib (approved):
Cool, thanks alot!

PR #815, opened 2025-01-21
Add Unicode-3.0 to accpeted licences

### LINKED ISSUE #814: License check fails due to Unicode-v3.0
Allow, or not allow?

According to https://www.unicode.org/policies/licensing_policy.html I think we should allow this license. What do you think @Niederb ?

haerdib:
ring licensing (3 in one file) confuses the cargo-about, so a workaround was added: https://embarkstudios.github.io/cargo-about/cli/generate/config.html

closes #814

### DIFF
--- a/about.toml
+++ b/about.toml
@@ -15,8 +15,7 @@ accepted = [
     "Unicode-DFS-2016",
     "Unicode-3.0",
     "NOASSERTION",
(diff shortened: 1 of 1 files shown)


Niederb (approved):
LGTM 👍

PR #835, opened 2025-02-05
Remove obsolete solochain-template-runtime dev-dep

haerdib:
No need to import the full solochain for a single Signature Type.


Niederb (approved):
LGTM

PR #846, opened 2025-04-28
Fix nightly build job

Niederb:
- Polkadot is working on adding support for decoding with a memory limit (`DecodeWithMemTracking`). Because of that we need additional trait bounds. See
  - https://github.com/paritytech/polkadot-sdk/issues/7360
- The `webpki-root-certs` changed the license. For me the license looks good See: 
  - https://github.com/rustls/webpki-roots/releases/tag/v%2F0.26.9
  - https://cdla.dev/permissive-2-0/
- In two tests we get fewer events. I could not find a reason for that, but I also did not search for long.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,7 +18,7 @@ version = "1.17.0"
 dependencies = [
  "ac-node-api",
  "ac-primitives",
- "frame-metadata",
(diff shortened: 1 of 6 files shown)


haerdib (approved):
Looks good, thank you very much for fixing the build!

PR #848, opened 2025-05-01
Lift dependencies to workspace and move api-client to its own package

clangenb:
The biggest motivation for this check is that encointer needs to depend on crates-io versions of the polkadot-sdk, and it is just so much easier for us to patch upstream if all the deps are defined in the workspace root only. Additionally, this gives you also small benefits like 

* ensuring that we use the same crate version throughout the whole workspace
* update deps at one place only

Follow-ups:
* This PR unveils that some of our CI stuff has only been running for the main crate `substrate-api-client` instead of all the crates because the root toml was both, a package toml and a workspace toml. As this also implies fixing of some clippy warnings, I thought it is better to move this to a follow-up issue #849.


.github/workflows/ci.yml
```diff
@@ -57,22 +57,22 @@ jobs:
...
-                cargo check --no-default-features --features std,
+                # Check build of stand-alone features.
+                cargo check --no-default-features -p substrate-api-client,
+                cargo check --no-default-features -p substrate-api-client --features sync-api,
+                cargo check --no-default-features -p substrate-api-client --features jsonrpsee-client,
+                cargo check --no-default-features -p substrate-api-client --features tungstenite-client,
+                cargo check --no-default-features -p substrate-api-client --features ws-client,
+                cargo check --no-default-features -p substrate-api-client --features staking-xt,
+                cargo check --no-default-features -p substrate-api-client --features contracts-xt,
+                cargo check --no-default-features -p substrate-api-client --features std,
```
clangenb, line 68: When I moved the `substreate-api-client` to its own pure package crate the default behaviour changed to executed these checks for the whole workspace. I put this change now to preserve the current behaviour as some CI checks actually fail. I decided that it is better to fix this in a follow up issue as there are also some clippy warnings.

.github/workflows/ci.yml
```diff
@@ -81,9 +81,9 @@ jobs:
                 cargo test --release -p ac-testing-sync,
 
                 # Clippy
-                cargo clippy -- -D warnings,
-                cargo clippy --no-default-features -- -D warnings,
-                cargo clippy --all-features --examples -- -D warnings,
+                cargo clippy -p substrate-api-client -- -D warnings,
+                cargo clippy -p substrate-api-client --no-default-features -- -D warnings,
+                cargo clippy -p substrate-api-client --all-features --examples -- -D warnings,
```
clangenb, line 86: Clippy is not happy for the whole workspace yet. 😭 #849

haerdib:
> Do you have any idea about the license check? I added the necessary license to accepted, but it still fails. I can locally reproduce this, but I don't really know what to do.

Try it by accepting only the GPL3.0-or-later, maybe this works. Cargo about is not very robust..

clangenb:
It complains that this does not exist... Shall we merge and create an issue?

haerdib:
> It complains that this does not exist... Shall we merge and create an issue?

Can you add the following accepted License? "GPL-3.0 WITH Classpath-exception-2.0" This should work.

clangenb:
Thanks, worked!

haerdib (approved):
Cool, I like this change. But let's wait with merging until @Niederb  has approved of it aswell. I'm not sure if he's currently working on something.

about.toml
```diff
@@ -12,7 +12,8 @@ accepted = [
     "Unicode-3.0",
     "NOASSERTION",
     "Zlib",
-    "GPL-3.0"
+    "GPL-3.0",
+    "GPL-3.0 WITH Classpath-exception-2.0"
```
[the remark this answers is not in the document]
clangenb: I don't think so. The exception makes it more liberal, so it should be fine.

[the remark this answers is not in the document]
Niederb: Yes, the Classpath-exception makes it more liberal, so this is fine 👍

.github/workflows/ci.yml
Niederb: Yeah, I also stumbled upon this a while ago. I did not investigate the root cause exactly then, but we should fix the clippy issues 👍

Niederb (approved):
I really like this change 👍 Less redundancies in the definition of the dependencies and versions

Cargo.toml
```diff
@@ -25,108 +14,76 @@ members = [
...
 
 # local deps
-ac-compose-macros = { path = "compose-macros", version = "1.17", default-features = false }
-ac-node-api = { path = "node-api", version = "1.17", default-features = false }
-ac-primitives = { path = "primitives", version = "1.17", default-features = false }
+ac-keystore = { default-features = false, path = "keystore", version = "1.17" }
+ac-compose-macros = { default-features = false, path = "compose-macros", version = "1.17" }
+ac-node-api = { default-features = false, path = "node-api", version = "1.17" }
+ac-primitives = { default-features = false, path = "primitives", version = "1.17" }
+substrate-api-client = { default-features = false, path = "api-client", version = "1.17" }
```
clangenb, line 84: @Niederb regarding your https://github.com/scs/substrate-api-client/pull/848#discussion_r2073530501. Yes the features are necessary, as we disable the default features here when we define the api-client as a workspace dependency here.

We could also do it the other way around as we need the default-features except for the no-std tests, but I decided to be consistent with the rest.

clangenb:
Cool thanks for the reviews. You can merge the PR at your convenience then.

Cargo.toml
Niederb: Oh, I see. Makes sense then 👍

clangenb: We could also explicitly put `default-features = true` in the async example to overwrite the workspace dependency setting. Not sure what is better though.

PR #868, opened 2025-09-18
CI: Change nightly build to weekly

haerdib:
As Parity and the api-client are much more stable, a weekly build for checking upstream changes should be enough.
@Niederb do you agree?

Thanks for the input @dhobi-scs!


Niederb:
Yeah, once a week should be sufficient 👍

PR #875, opened 2025-12-09
Prepare release of v1.19.0 (polkadot-stable2506-6)

Niederb:
DO NOT MERGE

- Switch polkadot dependencies to crates.io using the psvm tool
- Update internal version numbers


haerdib (commented):
LGTM - only the naming of the release and PR: Should be v1.19.0, or am I wrong?

